Issue #127, opened 2014-07-31
while (false) ++f shows ++f covered


hueniverse:

geek:
Fixed in latest https://github.com/hapijs/lab/blob/master/test/coverage.js#L143-L151

hueniverse:
We only we can also fix the broken line numbers in exceptions when coverage is on.

PR #141, opened 2014-08-07
Add test for while coverage

geek:
Cannot reproduce #127


cjihrig:
Can #127 be closed?

Marsup:
You're not testing the issue here, it has to be on the same line. I had it reproduced a while ago.

PR #220, opened 2014-09-26
node-growl notification support

csabapalfi:
Hi,

Would you be interested in merging Growl notification support? Happy to contribute that.

Mocha has that and I personally love it. I just like running tests with nodemon in the background and see notifications popping up.

Thanks,
Csaba


csabapalfi:
Instead of tap-growl now you can also use [lab-growl](https://github.com/csabapalfi/lab-growl) which understands the default lab reporter output.

PR #283, opened 2014-12-15
[CLI] Skip scripts which export null

yoitsro:
When a script inside the test folder exports null, lab throws at line 91 of `lib/cli.js`

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -88,6 +88,11 @@ internals.traverse = function (paths, options) {
             global._labScriptRun = false;
             file = Path.resolve(file);
             var pkg = require(file);
+
+            if (!pkg) {
+                return;
+            }
+
             if (pkg.lab &&
                 pkg.lab._root) {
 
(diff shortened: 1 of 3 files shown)


test/cli.js
```diff
@@ -537,6 +537,28 @@ describe('CLI', function () {
...
+            output += data;
+        });
+
+        cli.stderr.on('data', function (data) {
+
+            expect(data).to.not.exist();
+        });
+
+        cli.once('close', function (code, signal) {
+            expect(code).to.equal(0);
```
geek: @yoitsro please add a new line

lib/cli.js
```diff
@@ -88,6 +88,11 @@ internals.traverse = function (paths, options) {
             global._labScriptRun = false;
             file = Path.resolve(file);
             var pkg = require(file);
+
+            if (!pkg) {
```
geek: why export that ?

yoitsro: If you don't export it, will the script be executed?

geek: Ya, if you require the file it will execute the self-executing function like that

PR #290, opened 2015-01-07
make linter pluggable

smikes:
initial work: jslint runs but writes to stdout instead of returning data
update tests to use -L, -n to specify linter
style: remove dangling comma
spacing and intentation
rename linter output to "lint" from "eslint"
implement runner; add tests, fixtures for jslint
restore newlines
hoist helper functions
add jslint (w/version) as optional dependency

### DIFF
--- a/lib/linters/jslint/index.js
+++ b/lib/linters/jslint/index.js
@@ -10,6 +10,7 @@ var Nodejslint = require('jslint');
 var internals = {};
 
 var formatErrors = function (error) {
+
     return {
         line: error.line,
         severity: 'ERROR',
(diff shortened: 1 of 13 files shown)


hueniverse:
Isn't this a 5.2.0?

geek:
@hueniverse yes, hadn't got to reviewing fully to make that switch.  Thanks.

Marsup:
For what it's worth, I still believe a single switch is achievable :p

smikes:
@marsup My impression was that `bossy` did not want the logic needed to support a single switch.  Easier for me to make 1 PR than two connected ones; if you want to resurrect the other PR, that's fine by me.

@geek I agree with this as a minor version bump, 5.2.0 would be awesome :-)

lib/linters/jslint/index.js
```diff
@@ -0,0 +1,43 @@
...
+var formatErrors = function (error) {
+    return {
+        line: error.line,
+        severity: 'ERROR',
+        message: error.reason
+    };
+};
+
+var formatFile = function (file) {
+  return {
```
geek: Our styles a little different, 4 spaces and a newline after each function declaration.

smikes:
More like that?

geek:
@smikes perfect, thanks

I'll publish this later tonight.  Want to run through it all

PR #293, opened 2015-01-10
Closes #292

### LINKED ISSUE #292: coverage line chunks returns false string rather than false boolean
Here is an example when looking at the json output:

                    "74": {
                        "source": "   if (pail.scm && pail.scm.branch !== payload.scm.branch) {",
                        "chunks": [
                            {
                                "source": "   if ("
                            },
                            {
                                "source": "pail.scm",
                                "miss": "false"
                            },
                            {
                                "source": " && "
                            },
                            {
                                "source": "pail.scm.branch !== payload.scm.branch",
                                "miss": "never"
                            },
                            {
                                "source": ") {"
                            }
                        ],

Notice that the miss is "false" rather than just false.  I am writing a utllity that checks to see if this is a miss or not but since its string the check fails.  I would think it should be consistent with the other miss output which is just straight boolean false.

lloydbenson:
This seems to work but should be closely reviewed.  It seems like maybe this should just return true rather than 'never' but maybe there is a distinction between 'never' and miss: true that I am missing.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -446,7 +446,7 @@ internals.file = function (filename, data, options) {
                     }
 
                     isMiss = true;
-                    var issue = statement.hit.true ? 'true' : (statement.hit.false ? 'false' : 'never');
+                    var issue = statement.hit.true ? true : (statement.hit.false ? false : 'never');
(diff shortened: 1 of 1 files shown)


geek:
After talking with Lloyd, closing this issue

PR #308, opened 2015-02-16
 Remove Makefile and advice about it. Closes #284.

### LINKED ISSUE #284: Why are you running the lab command from a Makefile?
I just read the readme and I noticed you advise on writing the lab command in a Makefile. There's a bit of a problem here because then you have to specify the exact path to lab (in `node_modules/lab/bin/lab`). 

I usually write a script in the `package.json` scripts section, where you don't have to specify the full path to an installed package:

``` 
"scripts": {
    "test": "lab -t 100",
    "test-html": "lab -r html -o coverage.html"
  },
``` 

You can still have a Makefile (which I'm not sure is still necessary):
``` 
test-cov:
    npm test
test-cov-html:
    npm run test-html

.PHONY: test-cov test-cov-html
``` 

I know you're using it in most of hapijs packages, so a change in convention here is a global concern - I raise it here because it originated in here.

BTW, I generally don't  install a package globally; you can't make sure you have the right version installed and you have the npm scripts to make your life easier . So I would change the readme to reflect that - but it's your decision :smile:.

Oh, and another really minor issue that might arise by this convention is that if another dependent package (accidentally) installs lab in its `dependencies` (rather than `devDependencies`) lab might not reside in  `node_modules/lab/bin/lab` (but in `node_modules/otherPackage/node_modules/lab/bin/lab`) . But really, it's not an issue here. Sometimes, (on other packages) it might be - that's another reason why using an npm script is a good practice.

Marsup:


cjihrig:
+1 for getting rid of Makefiles. My only comment is that I'm not sure we need to worry about node 0.11 anymore. I think we were just testing it to make sure we were ready for the 0.12 release.

Marsup:
Node 0.11 has been out for a long time and I expect some companies to keep running it for a while, and it doesn't hurt having it, feel free to remove it though :)

PR #310, opened 2015-02-20
External reporter solution 1

joseph-norman:
First solution I outlined in the issue on external reporters (includes creating an external property which is changed in-source by the user. Might not be the best.


joseph-norman:
Will update coverage

Issue #306, opened 2015-02-10
Add src level transforms


ldesplat:
I hacked my current copy of lab to support react's JSX syntax and I was wondering if a generalized PR for applying source level transforms would be welcome?
This would allow you to write tests but also your application using ES6 syntax & beyond, support compile to javascript languages, and extensions like JSX without using a compile step before running tests.

To enable this, a transform option would have to be created and the following 2 files would have to support it:
- cli.js, loading test scripts in traverse function. Would have to overwrite require.extensions[] with new method & then return it once we're done with require of all test files.
- coverage.cli, in internals.instrument, would have to apply the transforms right after reading the contents of the file.

I am asking this question also because some transforms could change line numbers so coverage line numbers could be different. So, it is not very ideal if you are not careful what transforms you are applying to the files.

Any thoughts? I don't mind modifying Lab for my use case, but it would be nice to have such a feature in upstream.

Marsup:
I personally think it's out of the scope of lab and there are countless build systems, I'd rather have people deal with that on their own. Lab already supports sourcemaps which should help.

We'll see what other people think.

jedireza:
I was just talking to @geek about sourcemaps and jsx. I threw a spike together to demonstrate the `Coverage: 0.00% (0/0)` I see when trying using `node-jsx`. You can see it here:

 - https://github.com/jedireza/lab-jsx

Maybe [`node-jsx`](https://github.com/petehunt/node-jsx) isn't doing something right with sourcemaps.

@ldesplat are your modifications to lab on GitHub?

ldesplat:
@jedireza You can find them there: [https://github.com/ldesplat/lab/commit/da36d454e12b7de06d779570f6087df6297ee241?diff=split](ldesplat/lab)

I get the following output when running it in my project:
```
No global variable leaks detected
Coverage: 71.70% (45/159)
lib/components/Image.js missing coverage on line(s): 19, 20, ... 163, 164, 171, 172, 179, 180
Code coverage below threshold: 71.70 < 100
```

Marsup:
@jedireza I think node-jsx doesn't do anything at all about sourcemaps, which might be the problem here, and probably yours later :)

jedireza:
@ldesplat's changes do get coverage to display. https://github.com/jedireza/lab-jsx/tree/ldesplat

[...]

PR #311, opened 2015-02-21
Add live transform of source code before coverage analysis

ldesplat:
This pull request is in reference to Issue #306 .

This allows for anybody to add src level transforms. So in theory, one could write coffee code and be able to use Lab without a compile step beforehand.

Could I get someone to review these changes and let me know if I did anything wrong? I also need some comments on how to get the code coverage for my changes in coverage.js. I really don't understand how you guys are testing it.

I also know, I need to update documentation, but wanted to have some feedback before I invest in that too. Here is how you use it:

- Specify -T and a module name
`lab -c -T lab-babel ./test`

and this is lab-babel.js

```
var Babel = require('babel-core');

module.exports.extensions = ['.jsx'];
module.exports.transform = function (content) {

    var result = Babel.transform(content, { sourceMap: 'inline', ast: false});
    return result.code;
}
```

extensions tells lab what other type of files could be loaded in (.coffee, .es6, ...), and transform is the transform method.


Marsup:
If you're lost in coverage, do `npm run test-cov-html`, it's far more detailed.
As far as the parameter is concerned, I'd rather take a path or a simple name, this would allow you to `-T ./lab-babel.js` for a local file or `-T lab-label` for a node module.

lib/cli.js
```diff
@@ -277,5 +282,16 @@ internals.options = function () {
         options.ids = argv.id;
     }
 
+    if (options.transform) {
+        var transform = require(Path.resolve(options.transform));
+
+        if (Array.isArray(transform.extensions) && typeof transform.transform === 'function') {
+            options.transform = transform;
+        } else {
```
Marsup: Hapi guidelines say new line after }, even before else

Marsup:
Shouldn't you test that transformation actually works ? There are only failure tests.

ldesplat:
Thanks @Marsup for your comments

Yes, I did not have any comprehensive test coverage because I did not understand how to test it. Gotta wrap your head around testing lab with itself. Anyways, I have added it now and I am only missing 1 line of coverage. I truly don't know how to cover it.

`for (var i = 0, il = internals.transforms.length; i < il; ++i) {` in coverage.js line 27. Complains about i < il

Changed the API a bit to only transform the files you specify, and it now also transforms when you do not specify coverage option.

```
var Babel = require('babel-core');

module.exports = [
	{ext: '.jsx', transform: function (content) {

	    var result = Babel.transform(content, { sourceMap: 'inline', ast: false});
	    return result.code;
	}}
];
```

Anyways, if people are happy with the implementation and can advise me on getting the last line of coverage, then I can update the documentation as well.

ldesplat:
> As far as the parameter is concerned, I'd rather take a path or a simple name, this would allow you to -T ./lab-babel.js for a local file or -T lab-label for a node module.

It's just a path. I do Path.resolve. I don't want to get into resolving between a path and a module.

ldesplat:
I consider this PR ready unless there are any objections.

Now I am not making a great case for myself but another PR will have to submitted because I would like to pass an option to eslint (that is not possible to do over .eslintrc) and that is extensions in CLIEngine or --ext in cli. This allows eslint to read not just .js files but also other potential file extensions (ie. .js.es6 or .jsx) as it supports these features. Not an easy PR unfortunately since the linters are using ChildProcess.fork but I guess it could be passed as args.

Anyways, this is currently working great for my current project where I have some javascript code (with es6 features and jsx) running in both node and the browser and I am able to test it all with Lab pretty transparently.

README.md
```diff
@@ -43,6 +43,7 @@ global manipulation. Our goal with **lab** is to keep the execution engine as si
 - `-s`, `--silence` - silence test output, defaults to false.
 - `-S`, `--sourcemaps` - enables sourcemap support for stack traces and code coverage, disabled by default.
 - `-t`, `--threshold` - minimum code test coverage percentage (sets `-c`), defaults to 100%.
+- `-T`, `--transform` - javascript file that exports an array of objects ie. `[ { ext: ".js", transform: function (content) { ... } } ]`
```
geek: Please add a note that this can't be used alongside coverage

lib/cli.js
```diff
@@ -277,5 +286,17 @@ internals.options = function () {
         options.ids = argv.id;
     }
 
+    if (options.transform) {
```
geek: require `hoek` and assert on this option if it exists.

```js
Hoek.assert(Array.isArray(transform), 'transform module must export an array of objects {ext: ".js", transform: null or function (content)}');

options.transform = transform;
```

lib/coverage.js
```diff
@@ -16,13 +16,34 @@ var sourceMapSupport = require('source-map-support');
...
+        if (filename.indexOf(ext, filename.length - ext.length) !== -1) {
+            transform = element.transform;
+        }
+    });
+
+    if (transform) {
+        return transform(content);
+    }
+
+    return content;
```
geek: using a ternary here is fine instead of the above condition:

```
return (typeof transform === 'function') ? transform(content) : content;
```

lib/coverage.js
```diff
@@ -34,12 +55,27 @@ internals.prime = function () {
...
 };
 
-internals.prime();
-
 
 exports.instrument = function (options) {
 
     internals.patterns.unshift(internals.pattern(options));
+
+    if (options.transform) {
```
geek: Might be better off doing `Array.isArray(options.transform)` since you are calling `forEach` on it.

ldesplat:
Thank you @geek for the comments. I have made the changes.

geek:
@ldesplat awesome work, this will in 5.3.0, which will be published today

jedireza:
@ldesplat we're seeing inconsistencies with line coverage with html generated reports, the command line coverage seems accurate though. Have you noticed/tested that?

ldesplat:
@jedireza Expand your screen to the right. There are multiple columns on the table. You will see the right line numbers then.

But otherwise, yes the html report is not very nice. I had looked into generating a new report that assumes that transformed files are always the same (not 2 files into 1). This would simplify the report a lot and make it clear where the lines are and also could show the original code.

jedireza:
:pray: Thanks so much. It would be great if we could get the html highlighting on the right rows.

PR #316, opened 2015-03-02
Properly support sourcemaps when using the transform option

ldesplat:
Multiple changes:
 1) Fix the command line sourcemaps option to use alias so we can finally call it with -S and --sourcemaps as documented in README
 2) Pass filename to the exported transform method (2nd parameter) so it is backwards compatible. Sourcemaps do need that information to generate nice output.
 3) Sourcemaps are now transparently handled when using transforms. It was broken before but I did not realize as my transforms always gave me 1 to 1 mappings.

Issues:
- Testability: Unfortunately, to get code coverage for (3), I need to "install" an option to source-map-support module and I cannot do so in a way that can be undone or clean with the way things are done right now. Doing so, in tests, breaks Lab reporting.
- I can definitely write tests that will cover the the lines of code (by invoking the cli directly as in cli.js) but we will not get code coverage reports for that.

Question:
- When Lab crashes, is there a way to see the stack trace. It really likes to eat its own errors...

I apologize for the sourcemap support :(

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -4,9 +4,7 @@ var Fs = require('fs');
 
 // Declare internals
 
-var internals = {
-    fileCache: {}
-};
+var internals = {};
 
 internals.prime = function (extension, transform) {
 
@@ -17,11 +15,13 @@ internals.prime = function (extension, transform) {
 
         src = (typeof transform === 'function') ? transform(src, relativeFilename) : src;
 
-        internals.fileCache[relativeFilename] = src;
+        exports.fileCache[relativeFilename] = src;
 
         localModule._compile(src, filename);
     };
 };
+
+exports.fileCache = {};
 
 exports.install = function (settings) {
 
@@ -33,8 +33,8 @@ exports.install = function (settings) {
 
 exports.retrieveFile = function (path) {
 
-    if (internals.fileCache[path]) {
-        return internals.fileCache[path];
+    if (exports.fileCache[path]) {
+        return exports.fileCache[path];
     }
 
     var contents = null;
@@ -45,7 +45,7 @@ exports.retrieveFile = function (path) {
         contents = null;
     }
 
-    internals.fileCache[path] = contents;
+    exports.fileCache[path] = contents;
 
     return contents;
 };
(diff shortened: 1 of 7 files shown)


geek:
@ldesplat the code coverage will definitely be needed to merge this.  For the crash issue, can you create a PR with a test that demonstrates what you mean?

Thanks for doing all of the hard work on this feature, it will be very useful for a lot of people.

ldesplat:
100% code coverage and a feature that works in all of my projects.

The issue now, is just that I have not been able to write an automated Lab test that tests it a to z. I tried by doing a test in test/cli.js but none of the current tests try to do coverage analysis, and when I do I always get 0/0 coverage.

So, we get a nice 100% number but no meaningful tests. That's the best I could do...

geek:
@ldesplat nice work!

lib/transform.js
```diff
@@ -4,22 +4,48 @@ var Fs = require('fs');
...
 exports.install = function (settings) {
 
     settings.transform.forEach(function (transform) {
 
-        require.extensions[transform.ext] = function (module, filename) {
+        internals.prime(transform.ext, transform.transform);
+    });
+};
+
+exports.retrieveFile = function (path) {
```
geek: whats the reason to duplicate this function here and in coverage... would it be safe to have a single version used everywhere?

geek:
I don't see any, but do you think this PR has any breaking changes? @ldesplat

lib/transform.js
ldesplat: @geek It would be preferable yes. Now, that I look at it it seems it could be possible. I will have to move internals.transform in coverage.js and externalize it in transform.js. Maybe some other changes as well. I'll look into it.

I initially did it because I am storing all the transformed files in a cache (in internals.fileCache for transform.js and coverage.js), so that I can give it to source-map-support. Otherwise, source-map-support would read the file from the disk and it would not have the sourcemap available.

ldesplat:
There should no breaking changes. The API has changed slightly but it is adding a parameter to the transform method, so it's backwards compatible.

It also does give different output when using sourcemaps & transform since now it will be correct.. :)

ldesplat:
Rebased on top of Lab's new changes. Should be ready to go :)

PR #318, opened 2015-03-03
add debug option to show domain error stack

### LINKED ISSUE #294: More details from a domain exception
Recently I was getting an error:

```
Multiple callbacks or thrown errors received in test: ...
      at Domain.emit (events.js:95:17)
      at process._fatalException (node.js:249:27)
```

Which proved tricky to track down. I eventually opened up the local `./node_modules/lab/lib/runner.js` and added some debug info to the [domain error handler](https://github.com/hapijs/lab/blob/master/lib/runner.js#L390-L393):

```js
var onError = function (err) {

    console.log('*** ERR', err.stack);
    finish(err, 'error');
};
```

In my case, it turned out there was a `setTimeout` that wasn't being cleared before the test finished. So maybe that's why the stack wasn't shown.

Would it be helpful to add a `--debug` (for lack of a better name) option that printed the stack to help track down these kinds of issues?

jedireza:
In hopes to close #294, this PR adds the `-D`, `--debug` option. Change requests and feedback welcome.

### DIFF
--- a/test/runner.js
+++ b/test/runner.js
@@ -208,6 +208,29 @@ describe('Runner', function () {
         });
     });
 
+    it('debug domain error', function (done) {
+
+        var script = Lab.script();
+        script.experiment('test', function () {
+
+            script.test('a', function (finished) {
+
+                setImmediate(function () {
+
+                    throw new Error('throwing stack later');
+                });
+
+                finished();
+            });
+        });
+
+        Lab.execute(script, { D: true }, null, function (err, notebook) {
+
+            expect(notebook.errors.length).to.equal(1);
+            done();
+        });
+    });
+
     it('skips tests on failed before', function (done) {
 
         var steps = [];
--- a/test/runner.js
+++ b/test/runner.js
@@ -226,7 +226,7 @@ describe('Runner', function () {
 
         Lab.execute(script, { debug: true }, null, function (err, notebook) {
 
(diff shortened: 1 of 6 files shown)


test/runner.js
```diff
@@ -208,6 +208,29 @@ describe('Runner', function () {
...
+                setImmediate(function () {
+
+                    throw new Error('throwing stack later');
+                });
+
+                finished();
+            });
+        });
+
+        Lab.execute(script, { D: true }, null, function (err, notebook) {
```
jedireza: Something strange I noticed. If I use `debug` insead of `D` like this:

```js
Lab.execute(script, { debug: true }, null, function (err, notebook) {
```

I get the console artifacts printed to the command line.

test/runner.js
```diff
@@ -208,6 +208,29 @@ describe('Runner', function () {
...
+                setImmediate(function () {
+
+                    throw new Error('throwing stack later');
+                });
+
+                finished();
+            });
+        });
+
+        Lab.execute(script, { debug: true }, null, function (err, notebook) {
```
jedireza: There are some console artifacts from the `console.log` that was added to the `onError` function. Should I be pushing a new error to `state.report.errors` instead?

geek:
@jedireza thanks for the help, I restarted travis, it has a finicky test in v0.10

PR #320, opened 2015-03-03
Fixes #319: Instructs --output to create given directory recursively.

### LINKED ISSUE #319: --output should create given directory tree
Issuing the command `lab -a code -r html -o docs/coverage.html` should create the docs folder, if it doesn't exist; lab@5.2.1 fails with an exception:
```events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: ENOENT, open 'docs/coverage.html'
    at Error (native)```

### PR DESCRIPTION
**Pull request by jzorn**
This fixes the issue.

### DIFF
--- a/lib/reporters/index.js
+++ b/lib/reporters/index.js
@@ -30,7 +30,8 @@ exports.generate = function (options) {
     if (typeof options.output === 'string') {
         Mkdirp.sync(Path.dirname(options.output));
         dest = Fs.createWriteStream(options.output);
-    } else {
+    }
+    else {
         dest = options.output;
     }
 
--- a/lib/reporters/index.js
+++ b/lib/reporters/index.js
@@ -1,6 +1,8 @@
 // Load modules
 
 var Fs = require('fs');
+var Path = require('path');
+var Mkdirp = require('mkdirp');
 
 
 // Declare internals
@@ -24,7 +26,14 @@ exports.generate = function (options) {
     var Proto = internals.protos[options.reporter];
     var reporter = new Proto(options);
 
-    var dest = typeof options.output === 'string' ? Fs.createWriteStream(options.output) : options.output;
+    var dest;
+    if (typeof options.output === 'string') {
+      Mkdirp.sync(Path.dirname(options.output));
+      dest = Fs.createWriteStream(options.output);
+    } else {
+      dest = options.output;
+    }
+
     var output = '';
 
     reporter.report = function (text) {
--- a/lib/reporters/index.js
+++ b/lib/reporters/index.js
@@ -28,10 +28,10 @@ exports.generate = function (options) {
 
     var dest;
     if (typeof options.output === 'string') {
-      Mkdirp.sync(Path.dirname(options.output));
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by lloydbenson** [comment]
I can't comment if the maintainer will want this patch, but I know that the Mkdir package is problematic. There is a policy of no 0.x packages.  I'd try just strictly using fs.

**Comment by jzorn** [comment]
Oh I didn't know there was such a restriction. Especially since both `source-map-support` and `eslint` are 0.x packages. Anyways, if that's the case I'll find some time on the weekend and fix this in a different way.

File: lib/reporters/index.js [hunk 1]
```diff
@@ -24,7 +26,14 @@ exports.generate = function (options) {
     var Proto = internals.protos[options.reporter];
     var reporter = new Proto(options);
 
-    var dest = typeof options.output === 'string' ? Fs.createWriteStream(options.output) : options.output;
+    var dest;
+    if (typeof options.output === 'string') {
+      Mkdirp.sync(Path.dirname(options.output));
```
geek: we are 4 space indenters

geek:
Please add a test for this new functionality

geek:
@jzorn and @lloydbenson we try to stick with v1 modules when they are relied on by hapi in production.  Since lab is used for development it's sort of the exception on what it depends on.  Of course, it would be nice if a module that's ready for prime time is at least v1.

lloydbenson:
That's right my apologies.  Still seems like you could just use fs and not have the dependency but that's @geek 's call =)  Sorry for the diversion!

jzorn:
Sweet, sounds great - I'll fix the code style and add some tests and update the PR.

AdriVanHoudt:
Imo if you can do someting native without a dependency don't add another dependency

Marsup:
@AdriVanHoudt that goes against the very principle of npm, small modules that do few things but do it well.

AdriVanHoudt:
@Marsup I agree on that but if the module can be replaced by 5 lines of code it seems like more overhead. I looked at the code and it can't be done in 5 lines so lgtm :D 
There does seem to be an issue with infinite loops on windows...

jzorn:
Apparently recursively creating folders is not as easy as it appears. I strongly prefer code that is literally used in hundreds of projects over writing my own solutions, trying to cover every pitfall and corner case.

AdriVanHoudt:
@jzorn you're right, I thought it would be easier but it turns out not to be so using the module here is prefered

jzorn:
Great - I fixed the indent and will write a test and then check back with you.

lib/reporters/index.js
```diff
@@ -24,7 +26,14 @@ exports.generate = function (options) {
     var Proto = internals.protos[options.reporter];
     var reporter = new Proto(options);
 
-    var dest = typeof options.output === 'string' ? Fs.createWriteStream(options.output) : options.output;
+    var dest;
+    if (typeof options.output === 'string') {
+        Mkdirp.sync(Path.dirname(options.output));
+        dest = Fs.createWriteStream(options.output);
+    } else {
```
Marsup: else should be on his own line

PR #321, opened 2015-03-04
Allow --lint-options to be passed to configure linters

ldesplat:
Pass --lint-options with a JSON.parse(able) string. The options will be Hoek.merged with the configuration object in jslint or eslint. Now you can configure the linting program how you please and override or add to the defaults at your own peril.


geek:
@ldesplat you are on a roll!

PR #322, opened 2015-03-05
Fix domain not exited

Marsup:
Hi @geek,

Found out a possible fix for the bug we discussed on gitter.
I have a hard time figuring out a test for it but this works.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -370,6 +370,7 @@ internals.protect = function (item, state, callback) {
             return;
         }
 
+        domain.exit();
         isFirst = false;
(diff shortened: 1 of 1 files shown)


Marsup:
Well it doesn't fix my bug at work, but that's still something.
The clearTimeout has no effect, I still get the test timeout after the 1st error is thrown by a domain, I don't understand...

Marsup:
@geek FWIW maybe I was wrong on this one, see https://github.com/joyent/node/blob/master/lib/events.js#L134-L135.
It doesn't hurt calling it multiple times anyway.

PR #323, opened 2015-03-05
Fix EventEmitter bug

Marsup:
@geek I finally managed to reproduce the bug reliably !

So basically the explanation is the domain of the beforeEach captures the event emitter that's created, and if it crashes in the context of the test, the error is not caught by the `protect` of the test but by the one of beforeEach.

I have an idea how to fix it but that's gonna be weird : we should transfer the list of `domain.members` from the beforeEach to the new domain of the test.
What do you think ?

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -154,6 +154,10 @@ internals.executeExperiments = function (experiments, state, skip, callback) {
 
     Items.serial(experiments, function (experiment, nextExperiment) {
 
+        // Create a new domains context for this level of experiments, keep the old ones to restore them when finishing
+        var previousDomains = state.domains;
+        state.domains = [];
+
         var skipExperiment = skip || experiment.options.skip;
         var steps = [
             function (next) {
@@ -202,6 +206,9 @@ internals.executeExperiments = function (experiments, state, skip, callback) {
             item(next);
         },
         function (err, results) {
+
+            // Restore the domains we had before
+            state.domains = previousDomains;
 
             if (err) {
                 state.report.errors.push(err);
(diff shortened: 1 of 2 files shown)


geek:
Does the other PR (#322) fix this test?

Marsup:
Nope not at all, but I think it's still worth merging since we do have a path that doesn't exit the domain, or maybe we should just use domain.run, any reason why it's not the case?

Marsup:
So what do you think of my suggestion ?

geek:
@Marsup I wish domains wasn't so broken... I think what you are proposing sounds right, with the way domains cascade... give it a try

Marsup:
Honestly with this use case, I'm not even sure async-listener would help, I don't know it well enough yet.
I'll work in this PR to fix it if you don't mind so don't close it yet.

Marsup:
@geek so my solution wasn't possible, since we don't manually register EEs and there's no way to get a list of them, and anyway it would have been functionally wrong.

Here's a patch that **should** fix domains once and for all, but it's not pretty. I can't say I'm 100% sure we won't see other edge cases considering the complexity of domains.

What happens with this patch is, since we can't know the exact origin of the error (still looking into that), we'll make all the started tests depending on an EE crash.
Let's say you have a beforeEach with an EE, and tests [A, B, C], if an error occurs during B, A and B will get the error, not C, that means that if A passed, it will error as `Multiple callbacks or thrown errors received...`, which is probably right since A could have caused the error after it passed and then make B crash. This gets ugly really fast if you enable parallel testing, a single EE throwing will make all of your tests depending on it crash.

To compensate, I'm now bringing back the real call stacks, so even if all of your tests crash, you will at least see the origin.

Another issue that will need to be fixed is errors that happen once the test suite is finished, they just go silent, but honestly after this patch I just wanna die so it'll wait :)

geek:
@Marsup nicely done, quite the feat!

PR #324, opened 2015-03-06
Allow test files to have different extensions with transform

ldesplat:
1) Since we specify transform, our tests themselves could be transformed.
2) Document a pitfall, which is to not transform too many files so the user has to check for that in transform method.


README.md
```diff
@@ -207,11 +207,16 @@ To use source transforms, you must specify a file that tells Lab how to do the t
...
 
 module.exports = [
-  {ext: '.js', transform: function (content, filename) {
+    {ext: '.js', transform: function (content, filename) {
 
-    var result = Babel.transform(content, { sourceMap: 'inline', filename: filename, sourceFileName: filename });
-      return result.code;
-  }}
+        // Make sure to only transform your code or the dependencies you want
+        if (filename.indexOf('node_modules') <= -1) {
```
geek: I think this can be `===`

lib/cli.js
```diff
@@ -12,7 +12,9 @@ var Utils = require('./utils');
 
 // Declare internals
 
-var internals = {};
+var internals = {
+    pattern: new RegExp('.(js)$')
```
geek: generally we do `/.(js)$/`

lib/cli.js
```diff
@@ -309,9 +311,17 @@ internals.options = function () {
         var transform = require(Path.resolve(options.transform));
 
         Hoek.assert(Array.isArray(transform), 'transform module must export an array of objects {ext: ".js", transform: null or function (content, filename)}');
-
         options.transform = transform;
+
+        var includes = 'js|' + transform.map(internals.mapTransform).join('|');
```
gergoerdosi: This will create a wrong regex ( `.(js|)$` ) if the transform module exports an empty array. Also, the dot should be escaped, below and above too.

ldesplat:
Thank you so much @gergoerdosi for spotting those issues.

I have fixed the code review issues and file extensions can now include some crazy stuff :). Tested a whole bunch of special characters on Mac, quite interesting!

geek:
@ldesplat do we need any new tests for any of the edge cases you were considering with this change?

ldesplat:
don't merge

ldesplat:
Ok, I am sorry. All should be well now.

Thanks for telling me to add tests.
There was a bug where applying transforms with the no code coverage path was different than the coverage path. They are now using the exact same code and it is consistent behavior. This is covered in the 2 new tests.

I had to resort to way too many PRs to add this feature in. Next time I'll hold on a bit longer no matter my enthusiasm! Thanks for all your support.

AdriVanHoudt:
@ldesplat thanks for all the effort to add this feature!

Issue #332, opened 2015-03-23
using `espree` or `acorn` instead of esprima


tunnckoCore:
Hi again. When try to run koa tests it fails because esprima not support generators and many more features actually. So i tried to replace esprima with `espree` and it works without any changes, because it's almost 100% compatible with esprima - actually it is a fork.

```js
var tree = Esprima.parse(content, {
    loc: true,
    comment: true,
    range: true,
    ecmaFeatures: {
        generators: true
    }
});
```

**list of es6 features..** or https://github.com/eslint/espree
```js
    ecmaFeatures: {

        // enable parsing of arrow functions
        arrowFunctions: true,

        // enable parsing of let/const
        blockBindings: true,

        // enable parsing of destructured arrays and objects
        destructuring: true,

        // enable parsing of regular expression y flag
        regexYFlag: true,

        // enable parsing of regular expression u flag
        regexUFlag: true,

        // enable parsing of template strings
        templateStrings: true,

        // enable parsing of binary literals
        binaryLiterals: true,

        // enable parsing of ES6 octal literals
        octalLiterals: true,

        // enable parsing unicode code point escape sequences
        unicodeCodePointEscapes: true,

        // enable parsing of default parameters
        defaultParams: true,

        // enable parsing of rest parameters
        restParams: true,

        // enable parsing of for-of statement
        forOf: true,

        // enable parsing computed object literal properties
        objectLiteralComputedProperties: true,

        // enable parsing of shorthand object literal methods
        objectLiteralShorthandMethods: true,

        // enable parsing of shorthand object literal properties
        objectLiteralShorthandProperties: true,

        // Allow duplicate object literal properties (except '__proto__')
        objectLiteralDuplicateProperties: true,

        // enable parsing of generators/yield
        generators: true,

        // enable parsing spread operator
        spread: true,

        // enable parsing classes
        classes: true,

        // enable parsing of modules
        modules: true,

        // enable React JSX parsing
        jsx: true,

        // enable return in global scope
        globalReturn: true
    }
```

Marsup:
I'd vote for espree considering the way the project is lead.

tunnckoCore:
yea, looks great :+1:

geek:
Thoughts on esprima v2 ?

tunnckoCore:
```
2015-03-09: Version 2.1.0

      * Support ES6 class (issue 1001)
      * Support ES6 rest parameter (issue 1011)
[...]

PR #335, opened 2015-04-06
Replace esprima with espree

Josiah:
Provides support for ES6 features such as generators.

Suggested in #332

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -7,7 +7,7 @@
 
 var Fs = require('fs');
 var Path = require('path');
-var Esprima = require('esprima');
+var Espree = require('espree');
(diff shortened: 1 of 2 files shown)


Josiah:
Playing with this locally it looks like it's not a simple 'Drop it in and go'.

I've encountered issues with getting ES6 files to parse correctly using espree (specifically around yield statements). I'll keep looking into this, but this is not the right solution as is.

ldesplat:
Don't you need to provide options to the parse method like `{ ecmaFeatures: { generators: true } }` ?

Josiah:
I had that option in and it failed a different place. I'm going to have another look at this when I've got more time.

PR #339, opened 2015-04-08
Add ESLint rule for blank line at beginning of function scope

cjihrig:
The rule causes warnings if the blank line is missing. Testing turned up a couple places where the blank line was missing in lab.

### DIFF
--- a/lib/lint.js
+++ b/lib/lint.js
@@ -19,6 +19,7 @@ exports.lint = function (settings, callback) {
                                   settings['lint-options'] ? [ settings['lint-options'] ] : [],
                                   { cwd: settings.lintingPath });
     child.once('message', function (message) {
+
         child.kill();
         return callback(null, { lint: message });
     });
--- a/lib/linters/eslint/.eslintrc
+++ b/lib/linters/eslint/.eslintrc
@@ -1,4 +1,7 @@
 {
+    "plugins": [
+        "eslint-plugin-hapi"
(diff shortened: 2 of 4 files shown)


Marsup:
:+1: Shouldn't we create a single module for all hapi-related rules ? One module per rule seems overkill.

Marsup:
Also, wouldn't it be better to release it as an ESLint plugin ?

cjihrig:
This was the only rule that @geek brought to my attention. What else is missing/not natively supported by ESLint?

Marsup:
There were a few when I first implemented ESLint into lab, maybe it changed. I'll try to put up a list someday.

AdriVanHoudt:
@cjihrig I love the rule! But can you make it into a plugin so I can use it in other projects (without using cli params) (see https://github.com/eslint/eslint/issues/2180)

cjihrig:
I'll look into creating a plugin today.

AdriVanHoudt:
@cjihrig cool thanks!

cjihrig:
Converted to a plugin

AdriVanHoudt:
@cjihrig thanks I will definitely use this!

cjihrig:
:+1:

tielur:
@geek I'm not familiar with the release cycle for lab, but when do you think this 5.5.2 will be pushed to npm?

Issue #338, opened 2015-04-08
Source maps show wrong lines


romeovs:
I'm using `lab` for a small project and it's working very good.

I'm using running the test coverage with a custom `babel` transform (taken from the example
of the README):

```js
'use strict';
var Babel = require('babel');

module.exports = [
  {
    ext: '.js'
  , transform: function (content, filename) {
      // Make sure to only transform your code or the dependencies you want
      if (filename.indexOf('node_modules') === -1) {

        var opts  = {
          sourceMap: 'inline'
        , filename: filename
        , sourceFileName: filename
        , stage: 0
        , optional: [
            'runtime'
          ]
        };

        var result = Babel.transform(content, opts);
        return result.code;
      }

      return content;
    }
  }
];
```

The displaying of the uncovered lines seems off though, an seems to be shifted by 6 lines (as can be seen [here](https://coveralls.io/builds/2291010/source?filename=src%2Fimmutable.js)).

What can I do to fix this?

ldesplat:
I does not look like the Lab's lcov reporter takes the sourcemaps output into consideration like the console or html reporter.

It should be looking at line.originalLine amongst other things.

Marsup:
The files are also wrong in the console reporter since a few releases.

romeovs:
@ldesplat I also have the same problem with the `html` reporter.

ldesplat:
@romeovs Drag your browser page from the html report to the right. There are two extra columns that show the file it comes from and the real line number. I agree the html report is not optimized for the transform use case where there is a 1 to 1 mapping from source to transformed.

romeovs:
@ldesplat I do not see what you mean. This is what I see:
![screen shot 2015-04-09 at 01 04 27](https://cloud.githubusercontent.com/assets/1250185/7057279/6a79a01c-de54-11e4-8aca-25ab89b3d3a7.png)

Do you mean the rightmost column?

I'm talking about the fact that line 41 is marked for example, while there is clearly not much to test there.

ldesplat:
@romeovs Yes, that's what I meant. Indeed it highlights line 41 for example, but then it says it's actually line 36 on the right. So it is reject(err); that is not covered.

That's what I meant by the html reporter not being very good for this use case.

romeovs:
Ok maybe it would be nice to have an option to make the red lines over the original code? This is much more intuitive to me since to code shown is the original code, not the compiled code.

Also how should I interpret the following:
[...]

PR #341, opened 2015-04-10
Fix HTML reporter when using live-transformed code

ldesplat:
This will allow the HTML reporter (and others) to display the transformed code. The display now makes much more sense, and the right lines are highlighted.

@jedireza raised this earlier this week along.
@romeovs raised it too in issue #338 but this does not fix 338. Though it does fix the weird line numbers (line 1 being repeated) raised later on!

screen shot 2015-04-09 at 9 38 23 pm

Before, it would have highlighted an irrelevant line like so
screen shot 2015-04-09 at 9 40 48 pm

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -236,7 +236,8 @@ internals.instrument = function (filename) {
 
     // Store original source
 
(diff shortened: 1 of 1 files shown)


jedireza:
Thanks so much :pray:

Issue #79, opened 2014-04-17
Supporting promises


scottcorgan:
Does Lab support promise testing similar to Mocha?

k-sheth:
I am not sure if this made it into some release or not. If it did, would appreciate pointing me to a test case / doc that shows how it works. I have been yo-yo ing between mocha and lab because of domains/exceptions and promises.

thx

Traviskn:
I was having a bit of trouble figuring out how to get lab to work with promises.  When an assertion fails within a promise's then block, the error is unhandled and the test times out rather than reporting actual and expected values like normal.  The key to getting things working is to be sure to have a catch block at the end that passes the error to lab's done function - failed assertions will be reported as normal.  Simple example below:

```javascript
var Code = require('code');
var Lab = require('lab');
var lab = exports.lab = Lab.script();

import Promise from 'bluebird';


lab.experiment('math', function() {

  lab.test('assert within a promise', function (done) {

    new Promise(function(resolve, reject) {
      return resolve(2+2);
    })
    .then(function(four) {
      Code.expect(four).to.equal(3);
      done();
    })
    .catch(function(error) {
      // this makes sure any assertion errors are handled and reported by lab
      done(error);
    });
  });
});
```
This kind of solution may be obvious, but in case anyone else is struggling with promises in lab I thought I'd post this!

robkorv:
@Traviskn, great commend, I also was struggling with testing promises. An example would be a great addition to the README. A test even more.

robkorv:
There are some situation where @Traviskn example doesn't work. When you have an assert in `then()` or `catch()` and it doesn't trigger, the assert won't go off. You should add a custom error to `done()` in `then()` or `catch()` when you don't expect it to trigger.
```javascript
lab.test('expect an error from a promise', (done) => {

    return new Promise((resolve, reject) => {

        try {
            resolve(2);
        }
        catch (err) {
            reject(err);
        }
    }).then((result) => {

        console.log('resolved');
        done(new Error('promise should be rejected and caught'));
    }).catch((error) => {

        console.log('rejected, this does not trigger');
        Code.expect(error).to.exist();
        done(error);
    });
});
```
__complete example__
```javascript
'use strict';
const Code = require('code');
const Lab = require('lab');
const lab = exports.lab = Lab.script();



lab.experiment('promises resolving rejects', () => {

    // @Traviskn example, works as expected when resolved
    lab.test('failed assertion within a promise, resolve', (done) => {

[...]

PR #342, opened 2015-04-13
Handle Promises returned by tests

aliem:
Feature requested by issue #79

It does not expect a full fledged Promise, only a *thenable* since it is the most barebone Promise implementation. Every Promise is a thenable, not every thenable is a Promise, but a Promise can handle a thenable.

sample use case:

```js

lab.experiment('Thenable', function () {
    lab.test(function () {
        return new Promise(function () {
            // This Exception will be caught by the Promise 
            throw new Error('Meaningful Exception');
        });
    });
});

lab.experiment('Thenable', function () {
    lab.test(function () {
        return Promise.all([
            Promise.fulfill('one'),
            Promise.fulfill('two'),
            'three'
        ]);
    });
});


lab.experiment('Thenable', function () {
    lab.test(function () {
        return Promise.all([
            Promise.fulfill('one'),
            Promise.reject(new Error('Something went terribly wrong'))
        ]);
    });
});

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
   },
   "devDependencies": {
     "code": "1.x.x",
-    "lab-event-reporter": "1.x.x"
+    "lab-event-reporter": "1.x.x",
+    "promise": "^6.1.0"
   },
   "bin": {
     "lab": "./bin/lab"
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -442,10 +442,13 @@ internals.protect = function (item, state, callback) {
     setImmediate(function () {
 
         domain.enter();
-        item.fn.call(null, function (err) {
-
+        var fn = function (err) {
             finish(err, 'done');
-        });
+        };
+        var thenable = item.fn.call(null, fn);
+        if (thenable && typeof thenable.then === 'function') {
+            thenable.then(fn.bind(undefined, null), fn);
+        }
         domain.exit();
     });
 };
--- a/test/runner.js
+++ b/test/runner.js
@@ -439,6 +439,10 @@ describe('Runner', function () {
     });
 
     it('can handle thenables', function (done) {
+        if (typeof Promise === 'undefined') {
+          var Promise = require('promise');
+        }
+
         var script = Lab.script();
         script.experiment('test', function () {
 
--- a/test/runner.js
+++ b/test/runner.js
@@ -438,6 +438,27 @@ describe('Runner', function () {
         });
     });
 
+    it('can handle thenables', function (done) {
+
+        var script = Lab.script();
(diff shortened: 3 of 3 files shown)


Marsup:
Your test is dubious, Promise doesn't exist in node 0.10.

aliem:
Sorry for this.

I tested it on iojs, but it is definitely strange that the tests pass on 0.10. It should simply throw an error in the domain.

Marsup:
Don't close it, add commits to it.

aliem:
As it stands It is worthless. When I run tests it won't actually run the tests. My steps to test it are:

1. copy `lib` to `test_runner` on master
2. switch to `feature/thenables`
3. `ROOT_SPAWN=1 ./bin/_lab -r tap -fL -t 100 -m 3000`

But the runner only runs up to the test 158 of 178. I'll figure it out in the evening, for now I'll continue to use `.catch(done)`.

aliem:
Quick followup. Lab does require a ['done' argument](https://github.com/hapijs/lab/blob/master/lib/index.js#L219), on each test.

But it seems that errors are not reported while running Lab's tests:

```
[lab]$ node ./bin/_lab -r console -f -m 3000 -g 'then'
[lab]$ echo $?
0
```

Since tests require an argument the change as it is is not useful since we can still use the `catch` method of good Promise implementations as the last statement. Using this change Lab could support simple thenables and Promises without adding a lot of noise and boilerplate:

```js
lab.test('thenable', function (done) {
  return thenable();
});

lab.test('Promise', function (done) {
  return Promise.fulfill();
});
```

instead of
```js
lab.test('thenable', function (done) {
  thenable()
  .then(done.bind(undefined, null), done);
});
lab.test('Promise', function (done) {
  Promise.fulfill()
  .catch(done);
});
```

Marsup:
I don't think done should be there when using promises, both use case can be detected, thought not upfront like currently.

aliem:
Yes, definitely it can be detected but it does require to refactor `Lab.test` putting the exception in the runner.

aliem:
f3eeeba uses Espree to parse the function looking for a return statement if the test does not accept a callback.

I'm pretty sure it will add quite a bit of overhead, expecially with complicated tests

Marsup:
Not sure it's necessary, if there's no argument and no thenable returned, then you should error.

package.json
```diff
@@ -27,7 +27,8 @@
   },
   "devDependencies": {
     "code": "1.x.x",
-    "lab-event-reporter": "1.x.x"
+    "lab-event-reporter": "1.x.x",
+    "promise": "^6.1.0"
```
geek: make it `6.1.x`

PR #346, opened 2015-04-22
Add filename to compilation errors.

Marsup:
With the new espree, my compilation is failing. It probably won't solve it but at least it shows the file on which it's failing.


Marsup:
nvm, I don't want to spend time testing this since it would be highly dependent on espree failures, I'll my traces when it fails.

PR #347, opened 2015-04-23
Added two eslint rules to lab to enforce the hapi code style.

hofan41:
Force functions to be defined by assignment and not by declaration, as well as enforce spacing after function name.


hofan41:
Closes #348

geek:
@hofan41 thank you!

PR #350, opened 2015-04-25
Lint thresholds

Marsup:
On my way I un-ignored test files, it seems important that they go through linting as well.
Since we're out of short handles for CLI, well, none.


AdriVanHoudt:
:+1: I like that tests can fail on linting, will make PR's so much easier. But can't you add shortcuts like `-t` ?

AdriVanHoudt:
There is still b, q, j, k, u :P those don't really make sense but if timeout can have -m... 
Also I should not read PR's at 10am in the weekend :D @Marsup

geek:
@Marsup well done, this looks solid.

Issue #305, opened 2015-02-10
passing flags to node


jmonster:
I'm looking for a way to pass `--es_staging` to the underlying node process in order to enable ES6 features: https://iojs.org/en/es6.html

Is there an existing way to do this that I've overlooked?

Marsup:
What process ? Lab doesn't spawn child processes except for linting which you probably don't care.

jmonster:
`lab test/file` is certainly using an installed `node` process, no?

jmonster:
I see what you mean though... hmph.

Marsup:
The hashbang is just a shortcut, you can also run `node --es_staging ./node_modules/.bin/lab test/file`.

jmonster:
:+1: brain fart last night; thank you, sir :)

pkyeck:
for me
`node --harmony ./node_modules/.bin/lab`
works, but 
`node --harmony ./node_modules/.bin/lab -t 100`
throws an error, saying: 

```bash
$ node --harmony ./node_modules/.bin/lab -t 100
/Users/xxx/src/index.js:3
abCov._line('/Users/xxx/src/index.js',3);let Hapi = req
                                             ^^^^
SyntaxError: Unexpected identifier
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/xxx/node_modules/lab/lib/coverage.js:29:36)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/xxx/test/server.post.register.js:6:14)
    at Module._compile (module.js:460:26)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/xxx/node_modules/lab/lib/coverage.js:34:28)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at /Users/xxx/node_modules/lab/lib/cli.js:105:23
    at Array.forEach (native)
    at Object.internals.traverse (/Users/xxx/node_modules/lab/lib/cli.js:101:19)
    at Object.exports.run (/Users/xxx/node_modules/lab/lib/cli.js:50:29)
    at Object.<anonymous> (/Users/xxx/node_modules/lab/bin/lab:3:23)
[...]

PR #354, opened 2015-05-05
Fix coverage modifying "use strict" statements.

Marsup:
Related to #305.


Marsup:
@geek can we get this merged ?

PR #362, opened 2015-05-19
Default both linting thresholds to 0.

### LINKED ISSUE #361: Lint option -L should fail test if any linting problems are found
Not sure what's the story with all the --lint extra flags but the default should be the strict case.

Marsup:
Fixes #361.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ global manipulation. Our goal with **lab** is to keep the execution engine as si
 - `-I`, `--ignore` - ignore a list of globals for the leak detection (comma separated)
 - `-l`, `--leaks` - disables global variable leak detection.
 - `-L`, `--lint` - run linting rules using linter.  Disabled by default.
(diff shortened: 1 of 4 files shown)


AdriVanHoudt:
maybe add some info in the docs that when linting is enabled it will fail the tests if linting warnings/errors are found

Marsup:
Sort of done.

AdriVanHoudt:
hmm that only suggest that when not provided a value it defaults to 0, it should really say that enabling linting will result in enabling those options

Marsup:
That's implied by "defaults to".

AdriVanHoudt:
I know but it could be read as 'defaults to 0 if the option is used but no value is specified'. Maybe nitpicking but just my thought

Marsup:
We'll see if @geek wants a better wording, it's clear enough for me.

AdriVanHoudt:
:+1:

PR #367, opened 2015-05-26
added extra rule (to make wreck pass linting)

AdriVanHoudt:
prevents wreck build (https://travis-ci.org/hapijs/wreck/builds/64105510) from failing and the rule should only be used when supporting <IE8

### DIFF
--- a/lib/linters/eslint/.eslintrc
+++ b/lib/linters/eslint/.eslintrc
@@ -20,6 +20,7 @@
         "no-unused-expressions": 0,
         "no-regex-spaces": 0,
         "no-shadow": 0,
+        "no-catch-shadow": 0,
         "global-strict": 0,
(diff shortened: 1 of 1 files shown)


Marsup:
The goal is to fix the error, not disable the cause of it. -1 on this.

AdriVanHoudt:
docs of eslint specifically state to disable the rule when not supporting IE8 or less

Marsup:
Variable shadowing is usually bad, in a catch or not.

AdriVanHoudt:
ok, if you have a suggestion on how to fix it on Wreck I will do it there and close this

Marsup:
Rename the parameter ?

AdriVanHoudt:
also what @lloydbenson suggested, will do

PR #368, opened 2015-06-08
Enforce capitalization of imported module variable names

cjihrig:
Per the style guide:

> Use uppercase variable names for imported modules

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -8,7 +8,7 @@
 var Fs = require('fs');
 var Path = require('path');
 var Espree = require('espree');
-var sourceMapSupport = require('source-map-support');
+var SourceMapSupport = require('source-map-support');
 var Transform = require('./transform');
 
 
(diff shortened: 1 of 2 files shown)


lib/cli.js
```diff
@@ -102,7 +102,7 @@ internals.traverse = function (paths, options) {
 
             global._labScriptRun = false;
             file = Path.resolve(file);
-            var pkg = require(file);
+            var pkg = require(file);    // eslint-disable-line hapi/hapi-capitalize-modules
```
hueniverse: Can't we only enforce this rule on module global variables?

cjihrig: I'll look into it. It should be possible by inspecting parent nodes.

cjihrig:
@hueniverse I added an option to the rule to only target assignments in the module's global scope.

hueniverse:
@cjihrig that should be the default so we don't need all these exception comments above.

cjihrig:
Yep, it is now :-)

PR #371, opened 2015-06-10
Use no-shadow lint rule

### LINKED ISSUE #369: no-catch-shadow
https://jslinterrors.com/value-of-a-may-be-overwritten-in-ie8

Why do we care?!

geek:
Remove IE8 rule and use broader no-shadow with a warning

Fixes #369

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -63,9 +63,9 @@ internals.traverse = function (paths, options) {
             return path;
         }
 
-        Fs.readdirSync(path).forEach(function (fileName) {
+        Fs.readdirSync(path).forEach(function (filename) {
 
-            var file = Path.join(path, fileName);
+            var file = Path.join(path, filename);
             var stat = Fs.statSync(file);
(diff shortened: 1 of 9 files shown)


hueniverse:
What's with all the `forEach` replacements?

geek:
@hueniverse I was wondering what you would say about those :) 

We were using them elsewhere in the code, and I wanted to be a little more consistent and contain the variables in the scope a little more.

hueniverse:
Ok. Was just unrelated to the issue being fixed.

hueniverse:
What's with all the `done` -> `finished`?

geek:
In the tests that was to avoid shadowing `done`, however, it may be more appropriate as `next`... do you have a preference?

hueniverse:
`done` should be the actual test where `next` or `testDone` should be the code being tested.

geek:
@hueniverse I updated the test script tests to use `testDone` instead of `finished`, see what you think.

Issue #358, opened 2015-05-15
unable to test when employing template strings


jmonster:
I'm enjoying some ES6 fanciness, e.g:

```js
callback(null, `https://${host}${path}`, headers);
```

However, Lab can't process this:

```bash
> node ./node_modules/.bin/lab --timeout 5000 --lint --coverage

/Users/johnny/Documents/lanai/node_modules/lab/node_modules/espree/espree.js:5435
        throw e;
              ^
Error: Line 12: Unexpected token ILLEGAL
```

So I've altered my `package.json` to pass in the necessary `--use_strict` argument that is needed to unlock this feature:

```js
"scripts": {
    "test": "node --use_strict ./node_modules/.bin/lab --timeout 5000 --lint --coverage",
    "start": "node --use_strict index.js"
  }
```

But once I do this, I encounter a new problem:
```bash
> node --use_strict ./node_modules/.bin/lab --timeout 5000 --lint --coverage

/Users/johnny/Documents/lanai/node_modules/lab/lib/coverage.js:248
        __$$labCov = {
                   ^
ReferenceError: __$$labCov is not defined
```

Of course if I drop `--coverage`, everything runs to completion just fine.. thoughts?

jmonster:
Hey @geek thanks for taking a look at this, but I'm still having the issue -- although without any mention of `__$$labCov`:

```bash
uku ➤ npm test                                                                                                                                                  git:master*

> uku@0.0.1 test /Users/jdomino/Documents/heroku/uku
> lab --timeout 5000 --lint --coverage

/Users/jdomino/Documents/heroku/uku/node_modules/lab/node_modules/espree/espree.js:5435
        throw e;
              ^
Error: Line 12: Unexpected token ILLEGAL
```

If I simply remove all usage of template strings in my code [by commenting them out], this command runs to completion just fine.

geek:
@jmonster I think we will want to change our options when we call espree to be tolerant of parser errors.  I will log an issue

Marsup:
@geek This is probably more a matter of enabling espree ES6 feature here.

PR #386, opened 2015-06-25
enable code coverage support with ES6

jmonster:
features enabled
====
+ [x] arrowFunctions
+ [x] templateStrings
+ [x] blockBindings
+ [x] generators
+ [x] forOf
+ [x] binaryLiterals
+ [x] octalLiterals

There are additional options that could be enabled as well: https://github.com/eslint/espree#usage but I have elected to enable the ones I use most heavily and supported under iojs or node12; I'm unsure what, if any, negative consequence there may be to enabling them all by default.

This PR resolves issues raised at https://github.com/hapijs/lab/issues/358#issuecomment-103936194

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -208,7 +208,20 @@ internals.instrument = function (filename) {
 
     // Parse tree
 
-    var tree = Espree.parse(content, { loc: true, comment: true, range: true });
+    var tree = Espree.parse(content, {
+        loc: true,
+        comment: true,
+        range: true,
+        ecmaFeatures: {
+            blockBindings: true,
+            arrowFunctions: true,
+            templateStrings: true,
+            generators: true,
+            forOf: true,
+            binaryLiterals: true,
+            octalLiterals: true
+        }
+    });
 
     // Process comments
 


lib/coverage.js
```diff
@@ -208,7 +208,17 @@ internals.instrument = function (filename) {
...
-    var tree = Espree.parse(content, { loc: true, comment: true, range: true });
+    var tree = Espree.parse(content, {
+        loc: true,
+        comment: true,
+        range: true,
+        ecmaFeatures: {
+            blockBindings: true,
+            arrowFunctions: true,
+            templateStrings: true,
+            defaultParams: true
```
juhaelee: Can you add generators too?

jmonster: I'm down for adding them all -- whatsayyou @geek ?

geek:
I was talking with @cjihrig and he suggested we add a .labrc type file.  I think thats going to be useful for making lab more flexible to cases like this.  

Feel free to add generators to the list and I will merge in.

jmonster:
Updated PR to include the following:
+ [x] arrowFunctions
+ [x] templateStrings
+ [x] blockBindings
+ [x] generators
+ [x] forOf
+ [x] binaryLiterals
+ [x] octalLiterals

I selected these particular options since, according to an ES6 compatibility table, they have support in node 12 and iojs. Personally I've only been using the first 3 :)

PR #388, opened 2015-06-30
Enable multiple reporters

### LINKED ISSUE #285: Use reporters from json, support multiple reports
I have a use case where I want to get the nice native json output (which includes the code coverage) and from that json output, use the built-in reporters to put it in various formats.  The reason, I want to use the json file as the source, is because doing lab over and over for each format can potentially give different results (think unreliable downstream dependency) but is also obviously wasteful to re-test again.  It is nice having the 1 data source as the run result dataset, and then printing out whichever reporter you want (it would be nice to include console, html (coverage), junit, etc).  I don't care if i have to run each of these commands separately as they should be fast once the json file is complete.  I am open to really any ideas here.  The goal is to have 1 data set from a run, and use the internal reporters to format it based on that data set.

You could expose the functions so I could write another module or natively support this.  Let me know what you think and what the issues may be with doing this.

hofan41:
Closes #285

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -306,7 +306,7 @@ internals.options = function () {
     options.coverage = (options.coverage || options.threshold > 0 || options.reporter.indexOf('html') !== -1 || options.reporter.indexOf('lcov') !== -1 || options.reporter.indexOf('clover') !== -1);
 
     if (Array.isArray(options.reporter)) {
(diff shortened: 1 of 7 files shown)


hueniverse:
Probably worth a rebase... :-)

hofan41:
Sorry, I am no expert at git, I will figure out this rebase business and resubmit ;)

PR #389, opened 2015-07-01
Multiple reporters and multiple outputs

### LINKED ISSUE #285: Use reporters from json, support multiple reports

hofan41:
Closes #285


lib/reporters/index.js
```diff
@@ -21,6 +21,10 @@ internals.protos = {
 
 internals.requireReporter = function (reporter) {
 
+    if (Array.isArray(reporter)) {
+        return require('./multiple');
```
geek: Add a comment to the top of the file that indicates that multiple is required further down.  Just easier to look at the module section to find dependencies

```js
// ('./multiple') loaded below
```

lib/reporters/multiple.js
```diff
@@ -0,0 +1,91 @@
+// Load modules
+
+var Path = require('path');
+var Reporters = require('./index');
```
geek: Please move Reporters to be required last in the list
```
Path
Hoek
Reporters
```

We try to do:
```
Node core
3rd Party
Local
```

geek:
@hofan41 Please update the readme with information on how to use this enhancement

geek:
@hofan41 really well done, this is awesome!

hofan41:
Glad I could help! :thumbsup:

PR #392, opened 2015-07-02
Attach results to the reporter

geek:
Relates to #389

### DIFF
--- a/lib/reporters/multiple.js
+++ b/lib/reporters/multiple.js
@@ -4,9 +4,11 @@ var Path = require('path');
 var Hoek = require('hoek');
 var Reporters = require('./index');
 
+
 // Declare internals
 
 var internals = {};
+
 
 exports = module.exports = internals.Reporter = function (options) {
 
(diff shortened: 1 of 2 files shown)


geek:
@hofan41 what do you think of this change?

hofan41:
Looks good to me! Are there particular rules between when to use internals and when to use self/this?

geek:
It was more a concern of leaving the results on the static internals for the next time multiple is created... vs keeping the results with the instance.

hofan41:
Yeah I notice that now too. Good catch!

lib/reporters/multiple.js
```diff
@@ -4,10 +4,12 @@ var Path = require('path');
...
 
+
 // Declare internals
 
 var internals = {};
 
+
 exports = module.exports = internals.Reporter = function (options) {
 
     internals.reporters = [];
```
hofan41: Shouldn't we do the same with internals.reporters?

PR #395, opened 2015-07-05
Support filename pattern matching in CLI

### LINKED ISSUE #384: File name matching glob pattern
With nested folders for your code, it is sometimes nice to make your test files be along the production code, fx. `my-route-test.js` accompanying `my-route.js`. I would basically love to have an option for matching file patterns like this:

```bash
lab -P src/**/*-test.js
```

Is it on the roadmap?

geek:
Closes #384

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -12,9 +12,7 @@ var Utils = require('./utils');
 
 // Declare internals
 
-var internals = {
-    pattern: /\.(js)$/
-};
+var internals = {};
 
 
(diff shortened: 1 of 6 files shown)


lib/cli.js
```diff
@@ -12,9 +12,7 @@ var Utils = require('./utils');
 
 // Declare internals
 
-var internals = {
-    pattern: /\.(js)$/
```
AdriVanHoudt: so no more default?

PR #396, opened 2015-07-06
Test Reporters: Removing constraints and allowing multiple reporters

bapti:
- removed constraints to allow use of custom reporters either through `./local_mod` or node modules folder like `some_reporter`
- added multiple to be true to allow `-r html -r console` etc


geek:
@bapti there is a test case that fails in this situation, where we were erroring out before on an unknown reporter.  This will need to be updated before the PR would be accepted.  What other reporter were you wanting to use that you need this change?

PR #412, opened 2015-08-02
Update eslint to 1.0.x

gergoerdosi:
In eslint 1.0.0 all rules are off by default. So it's possible that we will need to add some other rules to .eslintrc or use `eslint:recommended`.

http://eslint.org/docs/user-guide/migrating-to-1.0.0


gergoerdosi:
The failing test seems to be unrelated to this change.

geek:
@gergoerdosi thanks, I am restarting the tests and checking on that error.

gergoerdosi:
@geek For some reason it fails locally on the master branch too. Maybe one of the dependencies got an update which broke something?

geek:
@gergoerdosi It looks like we have a bad assertion on https://github.com/hapijs/lab/blob/master/test/reporters.js#L831

Can you remove the `, 'Working right'` part... I know it's not really related to your PR, but it should resolve the breaking build.  

It looks like the `code` module fixed a bug that caused this issue to surface.

gergoerdosi:
@geek Pushed the change, build is passing now.

cjihrig:
With regards to `code` - we added an `options` argument to `equal()` for doing deep equal comparisons. The assertion in `lab` broke because it was passing an extra argument that was never supported.

AdriVanHoudt:
none of the new rules apply like `return-callback`? http://eslint.org/blog/2015/07/eslint-1.0.0-released/#new-rules

PR #420, opened 2015-08-04
Add command line options to override coverage path and excludes

### LINKED ISSUE #417: Not possible to ignore files from coverage
I want to have 100% coverage on the files that are in the `lib` directory. However I have files outside this directory too (server, configuration, etc.) that shouldn't have full coverage. I put together an example here: https://github.com/gergoerdosi/lab-coverage-example

When I run tests I get this output:

```
$ node ./node_modules/.bin/lab -a code -t 100 -L -v
Hello Route
  #hello
    ✔ 1) succeeds (26 ms)


1 tests complete
Test duration: 30 ms
Assertions count: 1 (verbosity: 1.00)
No global variable leaks detected
Coverage: 94.74% (1/19)
index.js missing coverage on line(s): 10
Code coverage below threshold: 94.74 < 100
```

Is there a way to tell lab to only check coverage on the content of the `lib` directory? I checked the `-g` and `-P` options, but looks like those are only for running tests. It is also possible to ignore files from linting using `. eslintignore`, but there doesn't seem to be anything for coverage.

I could add `// $lab:coverage:off$` and `// $lab:coverage:on$` to all the files outside `lib`, but there should be a better way.

gergoerdosi:
Closes #417

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ global manipulation. Our goal with **lab** is to keep the execution engine as si
 **lab** supports the following command line options:
 - `-a`, `--assert` - name of assert library to use.
 - `-c`, `--coverage` - enables code coverage analysis.
+- `--coverage-path` - sets code coverage path.
+- `--coverage-exclude` - sets code coverage excludes.
 - `-C`, `--colors` - enables or disables color output. Defaults to console capabilities.
 - `-d`, `--dry` - dry run. Skips all tests. Use with `-v` to generate a test catalog. Defaults to `false`.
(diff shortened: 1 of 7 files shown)


gergoerdosi:
@geek Here is my proposal to fix #417. Looking forward to your feedback.

geek:
@gergoerdos this looks great.  Can you also add details about the new arguments to the readme... after that I will merge and publish.

Thanks

gergoerdosi:
@geek Added the options to the README.

PR #421, opened 2015-08-09
Fix circular JSON for console reporter

Marsup:
Lab crashed silently with circular JSON in deep expectations.

### DIFF
--- a/lib/reporters/console.js
+++ b/lib/reporters/console.js
@@ -2,6 +2,7 @@
 
 var Tty = require('tty');
 var Diff = require('diff');
+var StringifySafe = require('json-stringify-safe');
 
 
 // Declare internals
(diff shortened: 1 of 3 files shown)


lib/reporters/console.js
```diff
@@ -2,6 +2,7 @@
 
 var Tty = require('tty');
 var Diff = require('diff');
+var StringifySafe = require('json-stringify-safe');
```
hueniverse: Use `Hoek.stringify()` instead https://github.com/hapijs/hoek#stringifyobj

Marsup: That's not what I want. Hoek only provides an error string instead, it's not because it's circular it cannot be diffed.

lib/reporters/console.js
```diff
@@ -143,8 +144,8 @@ internals.Reporter.prototype.end = function (notebook) {
             if (test.err.actual !== undefined &&
                 test.err.expected !== undefined) {
 
-                var actual = JSON.stringify(test.err.actual, null, 2);
-                var expected = JSON.stringify(test.err.expected, null, 2);
+                var actual = StringifySafe(test.err.actual, null, 2);
```
geek: Thoughts on falling back to StringifySafe if the normal stringify fails?  (https://github.com/isaacs/json-stringify-safe/issues/4)

I know we aren't super concerned with perf in testing, but it does seem reasonable to only use StringifySafe when needed.

Marsup: @geek stringify safe internally uses JSON.stringify with a custom replacer, not sure it's worth doing part of the job twice (worst case scenario), anyway would you mind updating your benchmarks with the newer version ? Note that it can still be hugely improved (see my whining here about it https://github.com/isaacs/json-stringify-safe/pull/12).

geek: Nice, thank you.

PR #428, opened 2015-08-19
Show invisible tokens on stringify

Marsup:
My team is constantly having problems with deep equals that show diffs without the hidden parts, it's very hard to debug, so this is one possible solution.


hueniverse:
Yeah, missing undefined is a problem. I am fine with this approach.

PR #431, opened 2015-09-02
Add docs for extending lab's linter

### LINKED ISSUE #372: .eslintrc files should build on default settings
So I have started a lab project and I disagree with one particular eslint setting. To change it setting-up a .eslintrc file seems to be most appropriate. However: when I add the file it replaces the `.eslintrc` settings of lab and falls back on all the standard eslint settings. This is not easy to follow/understand and annoying to fix.

It makes sense to me that it is done this way this because an .eslintrc file means that who-ever added this expects the standard .eslintrc behaviour. I do not see a good argument why the default of `lab` is not the default of `eslint`. There could be an option `-n hapi-eslint` for the convenience of hapi developers if necessary but then I would expect that `hapi-eslint` has a `.hapieslintrc` file that extends on the settings of `hapi-eslint`.

cjihrig:
Closes #372

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,16 @@ module.exports = [
 ];
 ```
 
+## Extending the linter
+
+**lab** uses a shareable [eslint](http://eslint.org/) config, and a plugin containing several **hapi** specific linting rules. If you want to extend the default linter you must:
+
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
:+1:

geek:
@cjihrig

PR #442, opened 2015-09-15
Update insecure dependencies (eslint, handlebars)

fhemberger:
handlebars@3.0.3 (also used in eslint @1.3.1) requires uglify-js@2.3.6, a
vulnerable version, see https://nodesecurity.io/advisories/module/uglify-js

This commit also addresses the changes introduced with the update of
both packages:

- eslint: Return the complete path of the linted file
  (see test/linters.js#30,51)
- handlebars: Depthed paths are now conditionally pushed on to the stack.
  (see https://github.com/wycats/handlebars.js/issues/1028)


geek:
@fhemberger thanks... happy to see eslint fixed that big bug on 1.4.0 :)

PR #444, opened 2015-09-15
closes #443

### LINKED ISSUE #443: add lint results to json reporter
It would be useful to have the lint object be a part of the json reporter.

lloydbenson:
This works, however I am finding it difficult to come up with a test case showing linting as linting is shutoff for the test directory.  Any thoughts on how to mock a test case?


geek:
@lloydbenson closing in favor of #445

PR #452, opened 2015-09-21
-V/--version flag to show version

danielb2:
CLI tools generally have a `--version` flag to print out version. Lab should too. 

I've caught myself running --version a number of times to see if I have the latest installed.


geek:
Fixes #449

Issue #453, opened 2015-09-25
Lint warning for unused variables except internals


hueniverse:

cjihrig:
I published v3.0.0 of `eslint-config-hapi`. I was about to make a PR to lab, but there are a lot of warnings. Most of them refer to unused function arguments, for example - https://github.com/hapijs/lab/blob/master/lib/reporters/lcov.js#L23

geek:
I think unused function arguments should be excluded from the check.  While an unused argument can be removed, they also help document the expected arguments.

hueniverse:
I agree. Unused function arguments are not a problem but the right way to write code. If you implement a callback you should include all the arguments expected as a way to document the interface.

PR #454, opened 2015-09-26
Update to eslint-config-hapi@3.x.x

cjihrig:
This turns on the ESLint rule `no-unused-vars`. All variables must be used, with the exception of `internals` and function arguments, otherwise a linting warning is created.


cjihrig:
Closes #453

PR #463, opened 2015-10-11
Fix error forwarding using active domain

tjmehta:
#### Example of Problem
```js
var Lab = require('lab')
var lab = exports.lab = Lab.script()
var before = lab.before
var beforeEach = lab.beforeEach
var through2 = require('through2')

var describe = lab.describe
var it = lab.it

describe('example', function () {
  var ctx
  describe('error emitted in beforeEach', function () {
    before(function (done) {
      ctx = {}
      ctx.stream = through2()
      done()
    })
    beforeEach(function (done) {
      emitErr()
      done()
    })

    it('should error', function (done) {
      done()
    })
  })

  describe('error emitted in test', function () {
    before(function (done) {
      ctx = {}
      ctx.stream = through2()
      done()
    })
    beforeEach(function (done) {
      done()
    })

    it('should error', function (done) {
      emitErr()
    })
  })
  function emitErr () {
    ctx.stream.emit('error')
  }
})
```

#### Lab incorrectly prints error locations
```bash

  xx

Test script errors:

Multiple callbacks or thrown errors received in test "Before example error emitted in beforeEach" (error): Uncaught, unspecified "error" event.
      at DestroyableTransform.emit (events.js:66:23)
      at emitErr (/Users/tjmehta/Developer/+lab/lab-test/lab-master/test-master.js:43:16)
      at /Users/tjmehta/Developer/+lab/lab-test/lab-master/test-master.js:19:7
      at processImmediate [as _immediateCallback] (timers.js:354:15)

Multiple callbacks or thrown errors received in test "Before each example error emitted in beforeEach" (done)
      at /Users/tjmehta/Developer/+lab/lab-test/lab-master/test-master.js:20:7
      at processImmediate [as _immediateCallback] (timers.js:354:15)

Multiple callbacks or thrown errors received in test "Before example error emitted in beforeEach" (error): Uncaught, unspecified "error" event.
      at DestroyableTransform.emit (events.js:66:23)
      at emitErr (/Users/tjmehta/Developer/+lab/lab-test/lab-master/test-master.js:43:16)
      at /Users/tjmehta/Developer/+lab/lab-test/lab-master/test-master.js:19:7
      at processImmediate [as _immediateCallback] (timers.js:354:15)

Multiple callbacks or thrown errors received in test "Before each example error emitted in test" (error): Uncaught, unspecified "error" event.
      at DestroyableTransform.emit (events.js:66:23)
[...]

PR #466, opened 2015-10-14
Improved emitter err handling using EmitterDomain, removed forwardError

tjmehta:
Using EmitterDomain we are able to determine the exact `before`, `beforeEach`, `test`, etc where an EventEmitter error occurs. Including both errors that occur in event handlers or emitted errors.

Continuing discussion from #463

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -1,6 +1,7 @@
 // Load modules
 
 var Domain = require('domain');
+var EmitterDomain = require('emitter-domain');
 var Items = require('items');
 var Reporters = require('./reporters');
 var Coverage = require('./coverage');
@@ -14,7 +15,10 @@ var Date = global.Date;
 var setTimeout = global.setTimeout;
 var clearTimeout = global.clearTimeout;
 var setImmediate = global.setImmediate;
-
+// emitter domains support more traceable event emitter error handling
+// they require wrapped event emitter methods to provide advanced behavior
+// wrap them here in case any tests modify EventEmitter
+EmitterDomain.wrapEventEmitterMethods();
 
 // Declare internals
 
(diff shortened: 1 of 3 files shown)


hueniverse:
Sorry but I don't like this. Domains are already fucked. This adds more complexity to a broken system. I admire the desire to work around domains and find solutions but the simplest we keep it, the better the chances of understanding why it fails when it does.

geek:
@tjmehta this looks like a reasonable solution.  But @hueniverse is right about the state of domains.  

Are there any enhancements you can make to the core node.js domain module to improve things?  I know it has a pending deprecation warning on it, but it may be worth trying to improve things at that level.

tjmehta:
Hello @hueniverse, first of all I just want to thank you for all of you open source contribution. It is really commendable what you have done and you are an inspiration to the community.

I actually agree with you here. EmitterDomain was kind've a thought experimentation and I actually agree with you I am not completely happy with it's current implementation. Currently there is only one side-effect I can think of, if any user is expecting an event emitter's `.domain` or `._events` to be what it normally is within an event handler it may not be what they expect. However, if the user is using their own domains, lab's emitterDomains will have no affect on user application logic.

The most I am unhappy with about how `lab` handles "eventEmitter errors" is that it prints the same error repeatedly for every `before`/ `beforeEach` use. IMO it is unnecessary to print the same stack 5 times like in the example shown in #463.

tjmehta:
Hello @geek, thank you for maintaining lab it is a great test runner and has been a pleasure to use.
> Are there any enhancements you can make to the core node.js domain module to improve things?

I am not sure of the performance memory implications of EmitterDomain, but I would assume that they are worse. I believe node core implemented event emitter error handling by domains as they did for efficiency.

I thought about contributing this back to node. but I saw the deprecated stamp on domains and this would be a major, breaking change. I also think that in application usage one would have more control over their emitter's error handling by using `domain.add`. Also, for handlers application code could `domain.bind` handlers to the active domain before attaching them.

With Lab, this is a bit harder to do. Lab runs application code it can end up handling event emitters deeply nested in application code.

I am mostly unhappy with about how Lab handles "eventEmitter errors" is that it prints the same error repeatedly for every `before`/ `beforeEach` use. IMO it is unnecessary to print the same stack 5 times like in the example shown in #463. Can we print a different error message "EventEmitter error.." along with the error's stack trace once? I have found the the stack trace's for EventEmitter errors that occur in lab to always be accurate and useful, but printing the same error n times misleads me into thinking that error has occurred n times.

geek:
@tjmehta I agree that its not good to print the same stack, but in the example in #463 only one of the stacks is duplicated.  I added an issue for this work: https://github.com/hapijs/lab/issues/467

PR #469, opened 2015-10-21
Add context as an alias to script.experiment.

pygeek:
Add context as an alias to script.experiment to improve organization of tests. This pattern is often seen in other BDD interfaces (see: mochajs.org and vowsjs.org).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -67,6 +67,7 @@ exports.script = function (options) {
     script.experiment.skip = internals.skip(script, 'experiment');
     script.experiment.only = internals.only(script, 'experiment');
     script.describe = script.experiment;
+    script.context = script.experiment;
     script.suite = script.experiment;
 
     script.test = internals.test.bind(script);


hueniverse:
Seems unnecessary.

pygeek:
I find this to be preferable to nesting experiments in some cases. Something I find useful in other test suites.

Here's another example (RSpec), and argument for context: http://lmws.net/describe-vs-context-in-rspec

Also, here's an example of a lab experiment with context: https://github.com/pygeek/abridge/blob/master/tests/integration/handlers/shorten.js

hueniverse:
It's just a alias, not any change in functionality of semantics.

pygeek:
Right. There is no change in functionality, it's purely a convention used in BDD. If you look at RSpec, for example, you'll notice that context *is* an alias to describe (https://github.com/rspec/rspec-core/blob/1f011150e965b12b3a6ea5123084bc20a9567984/lib/rspec/core/example_group.rb#L266-L276).

Also, here's a discussion on the topic of convention for RSpec: https://github.com/andreareginato/betterspecs/issues/3 .

pygeek:
Also, see https://mochajs.org/#bdd:

> The BDD interface provides describe(), context(), it(), before(), after(), beforeEach(), and afterEach().

> context() is just an alias for describe(), and behaves the same way; it just provides a way to keep tests > easier to read and organized."

geek:
I like the concept behind this change, and I know its not a major change either.  But for now, until there are more asks for it, I would rather it remain as a local shortcut in tests like below:


```js
// Test shortcuts

const describe = lab.describe;
const it = lab.it;
const context = lab.describe;

describe('#indexOf()', function () {
    context('when not present', function () {
        it('should not throw an error', function (done) {
...
```

PR #471, opened 2015-10-26
Do not add tracking code before "var" in "for(var x of xs)"

### LINKED ISSUE #470: for(var x of xs) is not handled correctly when coverage is turned on
I'm getting the error _Unexpected token var_ for this code when running with `-c` flag:

In file `test.js`:
```javascript
// require('lab') and other code deleted for clarity
require('./fail.js')
````

In file `fail.js`
```javascript
'use strict';
var actions=[1,2,3]
var rm={}
for (var action of actions) {
	rm[action] = true
}
```

If declaration of `action` is moved outside the for-loop it works
```javascript
var action
for (action of actions) {
```

Looks like ` && node.parent.type !== 'ForOfStatement'` needs to be added to https://github.com/hapijs/lab/blob/master/lib/coverage.js#L165 to get:
```javascript
if (trackedTypes.indexOf(node.type) !== -1 &&
    (node.type !== 'VariableDeclaration' || (node.parent.type !== 'ForStatement' && node.parent.type !== 'ForInStatement' && node.parent.type !== 'ForOfStatement')) &&
    (node.type !== 'ExpressionStatement' || node.expression.value !== 'use strict') &&
    node.parent.type !== 'LabeledStatement') {

    tracking.push(node.loc.start.line);
    node.set('global.__$$labCov._line(\'' + filename + '\',' + node.loc.start.line + ');' + node.source());
}
```

Found in lab version 6.2.0

HCanber:
Fixes #470 

Do not add tracking code before `var` in `for(var x of xs)`
This occurs when this is true:
```javascript
node.type == "VariableDeclaration"
node.parent.type == "ForOfStatement"
```

This PR includes new code in `partial.js` (as this was where I found `for(var x in xs)`) which is used in a test in `coverage.js`. 
I hope this was the correct place to add it.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -162,7 +162,7 @@ internals.instrument = function (filename) {
         ];
 
         if (trackedTypes.indexOf(node.type) !== -1 &&
-            (node.type !== 'VariableDeclaration' || (node.parent.type !== 'ForStatement' && node.parent.type !== 'ForInStatement')) &&
+            (node.type !== 'VariableDeclaration' || (node.parent.type !== 'ForStatement' && node.parent.type !== 'ForInStatement'  && node.parent.type !== 'ForOfStatement')) &&
             (node.type !== 'ExpressionStatement' || node.expression.value !== 'use strict') &&
             node.parent.type !== 'LabeledStatement') {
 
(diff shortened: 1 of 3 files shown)


Marsup:
LGTM. As for CI, maybe next lab should be node 4+.

geek:
@Marsup I think you are onto a good plan.  I will work on ironing out the details and see about making lab 7 be node 4+

PR #496, opened 2015-12-01
update to eslint-config-hapi@8.x.x

cjihrig:
This linter update enforces handling of errors in callbacks per https://github.com/hapijs/contrib/issues/65.


geek:
This can break existing builds, making it a major version change since its been a while since 7 was released.

Issue #493, opened 2015-11-24
Diff colors render diff text unreadable


ravinggenius:
When displaying a diff, the background colors are set. On my system (I'm using Solarized Light), the text color is already a not dark grey, so the text in the diff (arguably the most important part) is very difficult to read. Please explicitly set the foreground color anytime the background color is set. I've included a screenshot to demonstrate:

<img width="275" alt="screen shot 2015-11-17 at 21 10 52" src="https://cloud.githubusercontent.com/assets/21517/11383275/1ec4e646-92d5-11e5-8a5f-1a797149dbb7.png">

Marsup:
This looks like an easy one to fix, maybe a contributor could do it ?

alobodig:
For what it's worth, I use a dark color scheme and it's also difficult to read. I think setting the text to bright white whenever the background is set would fix readability in both situations.
![screen shot 2015-11-30 at 1 58 44 pm](https://cloud.githubusercontent.com/assets/1733093/11481720/16e20d86-976d-11e5-876a-9bb70fb74e9f.png)

thephilip:
What text color would everyone prefer?  Is black okay?  I'm new to contributing in general, but I would like to take a stab at this if that's okay with everyone else.  Might take me a couple of days because of work being busy.

alobodig:
Black is okay with me.

thephilip:
Working on it now...

thephilip:
Does this look okay?
![actual-expected](https://cloud.githubusercontent.com/assets/13875805/11634346/e0c10998-9cdd-11e5-81f8-dab8051fb64a.PNG)

I was able to test the colors by passing two new color codes into internals.colors.  Only problems are that the colors module is a bit finicky with multiple colors (specifically, I've had issues trying to implement the new colors with the built-in 'setTheme' method.  Also, I had to pass the new colors as a string but can re-factor internals.colors to accept and array of more than one value, but I would like to know what you all think.  I know passing a string in there is kinda hacky, but [again] I'm pretty new to this (sorry!).

```
    const codes = {
        'black': 0,
        'gray': 90,
        'red': 31,
        'green': 32,
        'yellow': 33,
        'magenta': 35,
        'redBg': 41,
        'greenBg': 42,
        'redBlack': '30;41',
        'greenBlack': '30;42'
    };
```

ravinggenius:
I think it's better, but the `actual` is still difficult to read. `expected` is okay though.

thephilip:
[...]

PR #500, opened 2015-12-07
fix diff text visibility for console reporter

thephilip:
Modified internals.color to accept an array of color codes and added new color codes to the internals.colors method as well.

See: [#493] (https://github.com/hapijs/lab/issues/493)

### DIFF
--- a/lib/reporters/console.js
+++ b/lib/reporters/console.js
@@ -339,7 +339,7 @@ internals.color = function (name, code, enabled) {
 
             return color + text + '\u001b[0m';
         };
-    } 
+    }
     else if (enabled) {
         const color = '\u001b[' + code + 'm';
         return function (text) {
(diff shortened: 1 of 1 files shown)


lib/reporters/console.js
```diff
@@ -341,6 +341,14 @@ internals.color = function (name, code, enabled) {
         };
     }
 
+    if (enabled && Array.isArray(code)) {
+        const color = '\u001b[' + code[0] + ';' + code[1] + 'm';
```
geek: doesn't look safe, you will want to check the length... might also want to do this condition first and else if enabled do the one on line 336

thephilip: I will do that.  Thank you for the suggestions.

thephilip:
@geek:  Besides not remembering code style requirements, I think I got the majority of what you asked for done.  I am considering adding the array length check in internals.colors instead of internals.color, but I'm unsure how it should be handled.  Should I output an error message/warning simply limit the array length, both, or something else?

PR #503, opened 2015-12-25
Fix eslint configuration file format support.

oparamo:
- Fix eslint configuration file checks.
- Use **.eslintrc.js** instead of **.eslintrc** as the default configuration file.
- Solves #504


oparamo:
Solves #504

geek:
@oparamo thank you for this, version 8.0.1 is now published with this fix

Issue #505, opened 2015-12-27
Coverage calculated wrong.


TomerAvni:
When running the coverage on "onPostHandler" method it show completely wrong results.

Here is the handler:
```
var Config = require('../config');
var Util = require('util');
var SessionAdapters = require('./common/session-adapters');
var SessionExtension = require('./session-ext');

exports.register = function (server, options, next) {
    var SessionStat = server.plugins['hapi-mongo-models'].SessionStat;

    server.ext('onPostHandler', function (request, reply) {
        var sessionId = SessionExtension.getSessionState(request);
        if (sessionId) {
            SessionStat.findById(sessionId, function (err, session) {
                if (err) {
                    return reply.continue();
                }

                if (request.auth.credentials) {
                    SessionAdapters.adapt(request.response.source, session, request.auth.credentials.roles);
                } else {
                    SessionAdapters.adapt(request.response.source, session, null);
                }

                return reply.continue();
            });
        } else {
            reply.continue();
        }
    });

    next();
};

exports.register.attributes = {
    name: 'session-adapters'
};
```
Now the the following tests are simple toggling weather the "sessionId" result exits or null:

```
var Lab = require('lab');
var Code = require('code');
var Path = require('path');
var Config = require('../../config');
var Manifest = require('../../manifest');
var Hapi = require('hapi');
var HapiAuth = require('hapi-auth-cookie');
var Proxyquire = require('proxyquire');
var AuthPlugin = require('../../server/auth');
var CookieAdmin = require('./fixtures/cookie-admin');



var lab = exports.lab = Lab.script();
var ModelsPlugin, server, stub;


lab.beforeEach(function (done) {

    stub = {
        SessionExtension: {},
        SessionAdapters: {},
        SessionStat: {}
    };


    var Extensions = Proxyquire('../../server/session-adt', {
        './session-ext': stub.SessionExtension,
        './common/session-adapters': stub.SessionAdapters
    });


    var proxy = {};
    proxy[Path.join(process.cwd(), './server/models/session-stat')] = stub.SessionStat;

    ModelsPlugin = {
        register: Proxyquire('hapi-mongo-models', proxy),
        options: Manifest.get('/plugins')['hapi-mongo-models']
    };

    var plugins = [ModelsPlugin, Extensions, HapiAuth, AuthPlugin];
    server = new Hapi.Server();
    server.connection({port: Config.get('/port/web')});
    server.register(plugins, function (err) {

        if (err) {
            return done(err);
        }

        done();
[...]

PR #510, opened 2016-01-11
Fix coverage when require cache is reset

tjmehta:
Require-mocking libraries (such as [proxyquire](https://github.com/thlorenz/proxyquire/issues/92)) and other modules may reset the require cache and enable requiring the same file multiple times.

### DIFF
--- a/test/coverage.js
+++ b/test/coverage.js
@@ -3,6 +3,7 @@
 // Load modules
 
 const Path = require('path');
+const Module = require('module')
 const Code = require('code');
 const _Lab = require('../test_runner');
 const Lab = require('../');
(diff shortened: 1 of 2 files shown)


tjmehta:
Fixes https://github.com/hapijs/lab/issues/505

test/coverage.js
```diff
@@ -232,4 +233,32 @@ describe('Coverage', () => {
...
+            done()
+        })
+
+        lab.after((done) => {
+
+            require.cache = Module._cache = cacheBackup; // restore require cache
+            done()
+        })
+
+        it('does not reset file coverage', function (done) {
```
geek: If you remove the lab.before and lab.after and put the code that changes the cache directly into the test, does it work?

test/coverage.js
```diff
@@ -232,4 +233,32 @@ describe('Coverage', () => {
             done();
         });
     });
+
+    describe('Clear require cache', () => {
+
+        const cacheBackup = require.cache; // backup require cache
+
+        lab.before((done) => {
```
geek: please move the setup and takedown into the test "does not reset file coverage"

test/coverage.js
tjmehta: yes it does, but if the test fails due to some runtime error the require cache will not restore.

test/coverage.js
tjmehta: Should I leave the describe?

tjmehta:
Removed before and after, lmk if anything else needs to be changed.

test/coverage.js
```diff
@@ -232,4 +233,21 @@ describe('Coverage', () => {
             done();
         });
     });
+
+    describe('Clear require cache', () => {
+
+        it('does not reset file coverage', function (done) {
```
geek: I ran this without the fix and it passes.  Can you create a test that fails without this fix

tjmehta: I see a mistake. Sorry about that.

tjmehta: Fixed.

gergoerdosi: Make it an arrow function.

tjmehta:
Changed function to arrow

geek:
@tjmehta thanks for the fix, looks good.  Going to publish as part of 8.0.2

PR #513, opened 2016-01-18
lint JS in Markdown files

maxbeatty:
When adding the CoffeeScript transform example in #512, I wanted to make sure it followed the style guide but that required running `eslint` against a temporary JS file. By happenstance, I came across [eslint-plugin-markdown](https://www.npmjs.com/package/eslint-plugin-markdown) yesterday which lints code fences in Markdown files. Now, whenever someone runs `npm test` or `npm run lint-md` the JS snippets in any `.md` files (primarily `README.md`) will be linted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ const lab = exports.lab = Lab.script();
 
 lab.test('returns true when 1 + 1 equals 2', (done) => {
 
-    Code.expect(1+1).to.equal(2);
+    Code.expect(1 + 1).to.equal(2);
     done();
 });
 ```
@@ -99,7 +99,7 @@ lab.experiment('math', () => {
 
     lab.test('returns true when 1 + 1 equals 2', (done) => {
 
-        Code.expect(1+1).to.equal(2);
+        Code.expect(1 + 1).to.equal(2);
         done();
     });
 });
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,8 +39,10 @@
   },
   "scripts": {
     "test": "./bin/_lab -fL -t 100 -m 3000",
+    "posttest": "npm run lint-md",
     "test-cov-html": "./bin/_lab -fL -r html -m 3000 -o coverage.html",
-    "lint": "./bin/_lab -d -f -L"
+    "lint": "./bin/_lab -d -f -L",
+    "lint-md": "eslint --config hapi --rule 'strict: 0, eol-last: 0' --plugin markdown --ext md  ."
```
maxbeatty: turned off `strict` since these are snippets and not whole files. IMO puts more focus on the code examples.

turned off `eol-last` because it wouldn't pass with one or two new lines in the code fence.

README.md
```diff
@@ -196,15 +210,13 @@ const lab = exports.lab = Lab.script();
 
 const suite = lab.suite;
 const test = lab.test;
-const before = lab.before;
```
maxbeatty: to comply with `no-unused-vars`, I deleted `before` and `after` here and used them in the BDD section to split the difference

geek:
@maxbeatty very nice.  Have you tried updating other hapi projects with this linting feature yet?

devinivy:
@maxbeatty @geek we do this with for the hapi tutorials, using the pre-1.0 version of that plugin.  I haven't seen it used anywhere else, but I'm a big fan of it :)

package.json
devinivy: In https://github.com/hapijs/hapijs.com/issues/217 we talked about what rules to allow/disallow, and we agreed on turning off hapi-scope-start.

PR #520, opened 2016-02-17
Fix coverage report for single-line functions and single-expression arrow functions

rluba:
It seems like lab does not correctly report coverage on various forms of short function declarations:

1. Any kind of function expression that’s written on a single line (see #274)
2. Single-expression arrow functions (eg. `(a, b) => a+b`), even if they span multiple lines.

This pull request adds a failing coverage test that demonstrates the problem. Unfortunately, I don’t know enough about lab’s AST transformations to be able to fix the problem.

PR #521, opened 2016-02-18
Fix coverage of trailing function declaration

rluba:
Declaring a function after a return statement causes lab to report the declaration as "not covered".

The new test demonstrates the problem: The old coverage code incorrectly reports lines 14 and 21 of `test/coverage/trailing-function-declarations.js` as "not covered". The fix in 2ba2d72 solves this problem by excluding function declarations from tracking.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -164,7 +164,6 @@ internals.instrument = function (filename) {
             'ReturnStatement',
             'ThrowStatement',
             'TryStatement',
-            'FunctionDeclaration',
             'IfStatement',
             'WhileStatement',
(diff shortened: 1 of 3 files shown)


rluba:
This PR will conflict with #520 because they both insert new tests at the same location in the test suite. Simply resolve it by arranging them in any order.

geek:
@rluba PR #520 is now merged if you want to rebase this

rluba:
@geek Thx, I’ve resolved the conflict.

geek:
@rluba well done, looks great.

Issue #524, opened 2016-02-19
Add a global `only` flag to improve TDD experience


rluba:
The current `only` flag only disables other experiments / tests within the same test file. From the documentation:
> This doesn't mark all other experiments and tests in a suite of scripts as skipped, instead it works within a single test script.

But during TDD in a codebase with many, many test files, I want solely run all tests _in a particular experiment_ very frequently while implementing a new feature — because always running all test files slows down each iteration unnecessarily. With mocha, I can simply [mark the corresponding test suite](http://mochajs.org/#exclusive-tests) with `.only(…)` during TDD development.

Lab has no such feature and the existing `only` flag is useless because I rarely have more than one experiment per test file.

If there’s interest in a global `only` flag similar to mocha’s `describe.only`, I’ll try to submit a PR to implement it. Any name suggestions? How about `{exclusive: true}`?

antony:
I for one would give a million upvotes for this. I've stopped using and removed Lab from all of our projects , and this is one of the core reasons. If it were fixed, I'd happily go back to it.

I would also vote for .only .exclusive etc as an alternative to using {only: true} et al. I like the fact I can use an options object in the test signature, but I also find it inconvenient. I find using it.only, describe.only much more preferable.

In fact I might suggest we change the purpose of 'only' (this isn't breaking, because people only use it temporarily, right?) to reflect the behaviour of every other test runner out there, and add a new flag, like 'here' or something which applies only within a single JS file.

Marsup:
The `.only`/`.skip` is supported since a very long time.

rluba:
@antony Sounds great. I’d prefer to change the meaning of `{only: true}` / `.only(` to match that of other frameworks, if that’s possible without alienating existing users.

@Marsup Really? I did not know that. Then we should probably add it to the "usage" documentation.

geek:
@rluba @antony another option is to use the index argument... this is how I accomplish what you are describing.  I think its a bit better than `only` because it requires no code changes.

Usually, I will do `lab -dv` to get a list of tests... then If I want to run a subset run `lab -i 4-7`, that will run tests numbered 4 through 7.  This also accepts a comma if you want to do a varying range.  `lab -i 4-7,9`  also, if you want to only run test 4 you can run `lab -i 4`

rluba:
[...]

PR #527, opened 2016-02-24
Global "only" (#524)

rluba:
I’ve implemented the global `only` flag as discussed in #524. Some of the test input files were copied & adapted from @Marsup’s original PR #116.

There are several topics that I’d like to hear everyone’s opinion on:

1. Currently, the runner simply ignores everything but the first `only` it encounters. I’d like to report an error instead but I don’t know how. There’s no mechanism for the runner to report general errors (only test errors). Every callback error is simply ignored (see [here](https://github.com/hapijs/lab/blob/fbd459675a857bf3ff9328969fda4fbc2fd7427a/lib/runner.js#L66) or [here](https://github.com/hapijs/lab/blob/fbd459675a857bf3ff9328969fda4fbc2fd7427a/lib/runner.js#L89)) and not forwarded to the reporters, so they can’t handle it. That’s why two [new](https://github.com/hapijs/lab/compare/master...rluba:feature/global_only#diff-583d616f4b01bf1fa153e71c0cef7de7R409) [tests](https://github.com/hapijs/lab/compare/master...rluba:feature/global_only#diff-583d616f4b01bf1fa153e71c0cef7de7R419) still fail.
2. The `ids` / `glob` feature currently uses [its own mechanism](https://github.com/hapijs/lab/blob/fbd459675a857bf3ff9328969fda4fbc2fd7427a/lib/runner.js#L299-L303) for ignoring tests. I’d like to merge this with the `skip` behavior so that tests ignored by the `ids` / `glob` parameters are also simply marked as "skipped" instead of not being part of the reports at all.
3. The console reporter now [prints the number of skipped tests](https://github.com/hapijs/lab/compare/master...rluba:feature/global_only#diff-425698733ef32d1551f3dbc61a9375dcR230). Shall we add this to other reporters as well?
4. I removed some 500 lines of boilerplate code from the cli tests by extracting a single function. I’ve also improved some of the assertions to be more precise. Any objections?
5. The before and after blocks are now consistently skipped for skipped tests. Previously, they were still executed for tests that were skipped individually (eg. using `test.skip(…)`).


lib/runner.js
```diff
@@ -124,6 +124,11 @@ exports.execute = function (scripts, options, reporter, callback) {
         return script._root;
     });
 
+    const onlyNodes = scripts.reduce((nodes, script) => nodes.concat(script._onlyNodes), []);
```
geek: Please wrap in `{`

```js
const onlyNodes = scripts.reduce((nodes, script) => {
  
    nodes.concat(script._onlyNodes);
}, []);
```

lib/runner.js
```diff
@@ -124,6 +124,11 @@ exports.execute = function (scripts, options, reporter, callback) {
...
+    if (onlyNodes.length) {
+        internals.skipEverythingExceptOnlyPath(scripts, onlyNodes[0]);
```
geek: consider rename, maybe `skipAllButOnly` ... needs to be shorter and simpler: `skipNononly` `filterSkip` `skipExperiments`

test/run_cli.js
```diff
@@ -0,0 +1,41 @@
...
+const ChildProcess = require('child_process');
+const Path = require('path');
+
+const labPath = Path.join(__dirname, '..', 'bin', '_lab');
+
+module.exports = (args, callback) => {
+
+    return new Promise((resolve, reject) => {
+
+        const cli = ChildProcess.spawn('node', [labPath, ...args]);
```
geek: no rest arguments supported yet

test/cli_only-skip/onlyRootExperiment.js
```diff
@@ -0,0 +1,267 @@
...
+
+        afterEach((done) =>  {
+
+            console.log('Should execute afterEach 2');
+            done();
+        });
+
+        it('returns true when 1 + 1 equals 2', (done) =>  {
+
+            expect(1+1).to.equal(2);
```
geek: please space infix operators `expect(1 + 1).to.equal(2);`

test/cli.js
```diff
@@ -9,7 +9,7 @@ const Code = require('code');
 const Lab = require('../');
 const Pkg = require('../package.json');
 const _Lab = require('../test_runner');
-
+const runCli = require('./run_cli');
```
geek: Please uppercase modules... this helps to make it easier to identify what is local and what is external

test/run_cli.js
rluba: It’s the spread operator, but you are right: I just saw that it’s [only supported in node 5, not in 4](https://kangax.github.io/compat-table/es6/).

test/run_cli.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const ChildProcess = require('child_process');
+const Path = require('path');
+
+const labPath = Path.join(__dirname, '..', 'bin', '_lab');
+
+module.exports = (args, callback) => {
+
+    return new Promise((resolve, reject) => {
```
geek: Please use the callback style instead of a Promise here

geek:
@rluba really well done!

1. Is the error when no `only` experiment or test is found?
2. I like the idea, but would the output show the skipped tests in the console reporter?  I am fine with making this change if we only show the skipped totals and not each skipped test in the reporter.  This is a feature that we use a lot in hapi and I don't want to make the reporter output too noisy when you only care about running a couple of tests in development. 
3. That would be great to add to the other reporters, but we can definitely do that as a separate PR to make it easier for this one to land.
4. Looks great... only change is to remove the rest params and to not use a Promise
5. Awesome!

test/run_cli.js
geek: Good to know, thanks.

lib/runner.js
Marsup: Isn't it more readable with a `Hoek.flatten(scripts.map((script) => script._onlyNodes))` ? Personal tastes...

lib/runner.js
```diff
@@ -147,7 +152,8 @@ exports.execute = function (scripts, options, reporter, callback) {
         },
         reporter: reporter,
         filters: filters,
-        options: settings
+        options: settings,
+        only: onlyNodes[0]
```
Marsup: I think that'll require a comment for further maintenance, it's not obvious why you picked the 1st.

lib/runner.js
```diff
@@ -164,6 +170,46 @@ exports.execute = function (scripts, options, reporter, callback) {
...
+    while (currentExperiment.parent) {
+        currentExperiment.parent.tests
+            .forEach(internals.enableSkip);
+        currentExperiment.parent.experiments
+            .filter(internals.not(currentExperiment))
+            .forEach(internals.enableSkip);
+        currentExperiment = currentExperiment.parent;
+    }
+
+    scripts.forEach((script) => {
```
Marsup: Why the sudden change of style ? filter + forEach would do as well.

test/run_cli.js
Marsup: Still behind a flag.

test/run_cli.js
```diff
@@ -0,0 +1,41 @@
...
+        cli.stderr.on('data', (data) => {
+
+            errorOutput += data;
+            combinedOutput += data;
+        });
+
+        cli.once('close', (code, signal) => {
+
+        	if (signal) {
+        		reject(new Promise(() => {
```
Marsup: Shouldn't ESLint pick this padding ?

test/cli.js
```diff
@@ -30,262 +30,125 @@ describe('CLI', () => {
...
+        runCli(['test/cli', '-m', '2000']).then((result) => {
 
-        cli.once('close', (code, signal) => {
-
-            expect(code).to.equal(0);
-            expect(signal).to.not.exist();
-            expect(output).to.contain('10 tests complete');
+            expect(result.errorOutput).to.equal('');
+            expect(result.code).to.equal(0);
+            expect(result.output).to.contain('9 tests complete');
```
Marsup: Is that normal we lost one ?

test/cli.js
```diff
@@ -325,465 +175,340 @@ describe('CLI', () => {
...
 
-        cli.once('close', (code, signal) => {
-
-            expect(code).to.equal(0);
-            expect(signal).to.not.exist();
-            expect(output).to.contain('1 tests complete');
-            done();
-        });
+            expect(result.errorOutput).to.equal('');
+            expect(result.output).to.contain('Should execute before 1');
```
Marsup: Contain supports arrays, you should do that instead.

test/run_cli.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const ChildProcess = require('child_process');
+const Path = require('path');
+
+const labPath = Path.join(__dirname, '..', 'bin', '_lab');
```
Marsup: Eran's gonna want his `internals` :)

test/run_cli.js
```diff
@@ -0,0 +1,41 @@
...
+
+module.exports = (args, callback) => {
+
+    return new Promise((resolve, reject) => {
+
+        const cli = ChildProcess.spawn('node', [labPath, ...args]);
+        let output = '';
```
Marsup: FWIW it's faster to push in arrays and join at the end, not sure it'll dramatically change test suite time though ;)

Marsup:
Done reviewing as well, great job.

test/run_cli.js
rluba: @Marsup It works in node 5 without a flag (tested on `v5.5.0`) as it’s not the "rest operator" but the "spread operator". (I find it pretty confusing that they look the same, have similar semantics but have completely different platform support.)

test/run_cli.js
rluba: That’s [verbatim from the original code](https://github.com/hapijs/lab/blob/b56cac853f9a5fee7e7f0b41d1f9ad89d592ae22/test/cli.js#L29) but I can change it.

rluba:
@geek

> Is the error when no only experiment or test is found?

No, no – that would break pretty much every current test case :-) I meant that I’d like to throw an error if there’s more than one `only` (doesn’t matter if it’s on experiments or tests).

I had this code in place, but there’s simple no place that would handle or output the error:

```JavaScript
    // (existing line)
    const onlyNodes = scripts.reduce((nodes, script) => nodes.concat(script._onlyNodes), []);
    // What I’d like to add:
    if (onlyNodes.length > 1) {
        const paths = onlyNodes.map((onlyNode) => {

            if (onlyNode.test) {
                return `Test: ${onlyNode.path}`;
            }
            return `Experiment: ${onlyNode.path}`;
        });
        // TODO: Report error somehow?
        return callback(new Error('Multiple tests are marked as "only":\n\t' + paths.join('\n\t')));
    }
```

> I like the idea, but would the output show the skipped tests in the console reporter?

Only if you use `-v`, otherwise you just see `(n skipped)` (see for example [this test](https://github.com/hapijs/lab/pull/527/files#diff-583d616f4b01bf1fa153e71c0cef7de7R345)).

lib/runner.js
rluba: I’ll use @Marsup’s version.

test/cli_only-skip/onlyRootExperiment.js
rluba: Sure, I just copied that from #116.

geek:
@rluba one option is to show the error in a similar way to https://github.com/hapijs/lab/blob/master/lib/cli.js#L116-L118  

I would like to keep it as is in regards to `-i` using skipped tests.  I tend to use `lab -v -i 44-50` or similar, and don't want to see a list of the skipped tests in that case.

test/run_cli.js
rluba: @Marsup You mean "no empty line"? Nope, it doesn’t. Only complains for functions, not for conditional blocks.

test/cli.js
rluba: Didn't know that, thx. Normally I use a [completely different assertion library](https://github.com/rluba/hamjest) because it provides far superior error messages.

test/cli.js
rluba: @Marsup Sharp 👀! [I had to delete the old "local only" test file](https://github.com/rluba/lab/commit/1dcd68ea7afe3bd29b9b200811bb26c8db071ba7) because it would have broken the "whole directory" test runs once it’s global.

lib/runner.js
rluba: `skipAllButOnly` it is!

rluba:
@geek Thanks, I’ll try that. It will produce `stdout` output instead of `stderr`, but better than nothing.

rluba:
@geek But the `process.exit` means that I can never cover that line according to lab: It’s covered in the "cli" tests but those don’t count against coverage because they run in a subprocess. Any suggestions?

lib/runner.js
rluba: @Marsup I hope the new version clarifies that. If not, let me know!

rluba:
So, I _think_ I’ve changed everything you’ve mentioned. Let me know if there’s something else.

test/run_cli.js
```diff
@@ -0,0 +1,38 @@
...
+    cli.stderr.on('data', (data) => {
+
+        errorOutput += data;
+        combinedOutput += data;
+    });
+
+    cli.once('close', (code, signal) => {
+
+    	if (signal) {
+    		callback(new Error('Unexpected signal: ' + signal));
```
Marsup: This is 8 spaces instead of 4, ESLint is missing something, unless I am...

geek: Its worse than you thought: tabs!

geek: looks like the new eslint ignore pattern needs to change :/  I got it

Issue #536, opened 2016-02-27
Support promises in tests and setup / teardown


rluba:
I know that this had been done in #342 and undone in #419 due to the complexity and overhead — but I believe that it can be implemented far simpler than #342 and without a performance penalty for those that don’t want to use promises.

It would remove a lot of error-prone boilerplate code for those of us that _do_ want to use promises.

Currently, I have to write:

```JavaScript

it('some test', (done) => {

    someAsyncCode()
        .then(() => {
             someAssertion();
             done();
        })
        .catch(done);
});
```

instead of just

```JavaScript

it('some test', () => {

    return someAsyncCode()
        .then(() => {
             someAssertion();
        });
});
```

If you are open to supporting the latter, I can provide a PR that will add this feature.

Marsup:
:+1: I also think this doesn't need to be that complex but since I don't have time to pick it up, you should, @geek seems open to that.

rluba:
Here you go!

PR #537, opened 2016-02-27
Feature: promises

rluba:
Implements #536 and fixes two bugs related to naming tests and setup / teardown blocks at the root level (outside any experiment).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -50,7 +50,8 @@ exports.script = function (options) {
         _current: {
             experiments: [],
             tests: [],
-            options: {}
+            options: {},
+            title: 'script'
         },
         _titles: [],
         _path: [],
@@ -210,7 +211,7 @@ internals.test = function (title /*, options, fn */) {
 
     const test = {
         path: this._path,
-        title: this._titles.join(' ') + ' ' + title,
+        title: (this._path.length ? this._titles.join(' ') + ' ' : '') + title,
         relativeTitle: title,
         fn: fn,
         options: settings
--- a/lib/index.js
+++ b/lib/index.js
@@ -134,7 +134,7 @@ internals.before = function (/* options, */ fn) {
     const options = arguments.length === 2 ? arguments[0] : {};
     fn = arguments.length === 2 ? arguments[1] : fn;
 
-    Hoek.assert(fn && fn.length === 1, 'Function for before in "' + this._current.title + '" should take exactly one argument');
+    Hoek.assert(fn, `before in "${this._current.title}" requires a function argument`);
 
     const before = {
         title: 'Before ' + this._titles.join(' '),
@@ -152,7 +152,7 @@ internals.after = function (/* options, */ fn) {
     const options = arguments.length === 2 ? arguments[0] : {};
     fn = arguments.length === 2 ? arguments[1] : fn;
 
-    Hoek.assert(fn && fn.length === 1, 'Function for after in "' + this._current.title + '" should take exactly one argument');
+    Hoek.assert(fn, `after in "${this._current.title}" requires a function argument`);
 
     const after = {
         title: 'After ' + this._titles.join(' '),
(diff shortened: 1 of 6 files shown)


lib/runner.js
```diff
@@ -564,10 +564,16 @@ internals.protect = function (item, state, callback) {
     setImmediate(() => {
 
         domain.enter();
-        item.fn.call(null, (err) => {
+        const itemResult = item.fn.call(null, (err) => {
 
             finish(err, 'done');
         });
+        if (itemResult && itemResult.then instanceof Function) {
+            itemResult.then(() => finish(null, 'done'), (err) => finish(err, 'done'));
```
Marsup: 2nd arg is only used in case of error.

rluba: Thx

lib/runner.js
```diff
@@ -564,10 +564,16 @@ internals.protect = function (item, state, callback) {
     setImmediate(() => {
 
         domain.enter();
-        item.fn.call(null, (err) => {
+        const itemResult = item.fn.call(null, (err) => {
 
             finish(err, 'done');
         });
+        if (itemResult && itemResult.then instanceof Function) {
+            itemResult.then(() => finish(null), (err) => finish(err, 'done'));
```
Marsup: Thought it was obvious but you don't really need null either, you could just .then(finish ;)

rluba: I thought you might mean that – but that would be incorrect. If the promise resolves with a value, `then(finish)` would pass the resolve value to `finish`, which would interpret it as an error!

Marsup:
You'll need tests with multiple `it` to check promises are not capturing anything.
I know the code should work but it's worth testing to avoid regressions.

rluba:
Seems like travis hit a timeout on 6c7df4a.

rluba:
>  to check promises are not capturing anything.

@Marsup, what do you mean by that?

Marsup:
Promises wrap everything in a try catch, so if finish failed synchronously the promise inside your test would catch it and you'd never know. It's prevented by the setImmediate of finish.

lib/runner.js
Marsup: Correct, nvm then.

rluba:
You’re right. Proper promise libraries like Bluebird have [built-in protection against such mistakes](http://bluebirdjs.com/docs/features.html#surfacing-unhandled-errors) but the native `Promise` does not.

Any suggestions on how to test for that without pulling in Bluebird or something similar?

Marsup:
For lab's small scope I'd suggest just testing that synchronous throws don't leak into any previous before/after/it.

rluba:
Sorry for the sloppy commits. I accidentally pushed my test of native promise behavior on unhandled rejections. Removed it in 411dc96.

lib/index.js
```diff
@@ -206,15 +207,11 @@ internals.test = function (title /*, options, fn */) {
...
-    if (fn) {
-        Hoek.assert(fn.length === 1, 'Function for test "' + title + '" should take exactly one argument');
-    }
-
     const settings = Utils.mergeOptions(this._current.options, options, ['only']);
 
     const test = {
         path: this._path,
-        title: this._titles.join(' ') + ' ' + title,
+        title: (this._path.length ? this._titles.join(' ') + ' ' : '') + title,
```
geek: is this meant to be `this_path.length` or `_this.titles.length` ?

rluba: It’s equivalent, these arrays alway have the same length. I can change it, if you think it’s confusing.

Marsup: Isn't it easier to just do `this._titles.concat(title).join(' ')` ?

rluba: Done.

Marsup:
Maybe we should increase the timeout on this test, it keeps failing on travis.

rluba:
@Marsup Not caused by this PR, but I can piggyback a fix.

rluba:
The JSON reporter test takes about a second on my machine, so I can understand that travis surpasses the 3s mark frequently.

test/reporters.js
```diff
@@ -1116,7 +1116,6 @@ describe('Reporter', () => {
                 expect(result.lint[0].filename).to.exist();
                 expect(result.lint[0].errors).to.exist();
                 done();
-                done = function () {};
```
Marsup: It's been done on purpose by https://github.com/hapijs/lab/commit/3e14d48960b5ca69e7a5f4c3e482090af7f37a9e. @geek was that a mistake ?

rluba: I know, but I see no way for it to be relevant unless there’s a bug in `Lab.report` that causes it to call its callback more than once.

And **if** such a case exists in `Lab.report`, I’d prefer this test to fail violently instead of shadowing the problem. Or did I miss something?

geek: thanks, removing the setting of done is a good move

PR #541, opened 2016-03-08
Allow stdout to flush

Marsup:
Because of `process.exit`, the output doesn't have the time to flush so you get a partial report (many issues about that in the node project).

This fixes it but I don't really have time to find how to cover those lines, I thought it already would be, so if anyone can look into it I'd be grateful.

### DIFF
--- a/lib/reporters/index.js
+++ b/lib/reporters/index.js
@@ -103,6 +103,11 @@ exports.generate = function (options) {
                     return callback(null, code, output);
                 }
 
+                if (dest === process.stdout) {
+                    // One last empty write to make sure we flushed the stdout buffer completely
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
why not just always flush it?

Marsup:
Because it's only valid if called from L115, otherwise `end` should take care of it.

geek:
@Marsup thanks, I will merge and get it covered.

Marsup:
BTW, for posterity, it would have been cleaner to check `needDrain` and listen to `drain` event, but those are internals, that's why I didn't rely on it.

AdriVanHoudt:
ah missed that!

PR #544, opened 2016-03-15
Feature/silent skips

rluba:
I’ve implemented a `silent-skips` option to reduce the noise when using `.only` in large test suites. I hope you like it.

With this in place, we could also remove the special case for `ids` / `glob` (see [bullet 2 in my `only` PR](https://github.com/hapijs/lab/pull/527#issue-136092402) and [@geek’s response](https://github.com/hapijs/lab/pull/527#issuecomment-188358276)) since the "noisy output" can now be prevented by adding `-k` when using `ids` / `glob`.

This PR also fixes `todo` tests being reported as `complete` at the end – even though they are displayed as skipped (`-`) in the output itself. I caught this by accident when writing the test for the new feature.

`master` currently contains some uncovered lines from #541 so `npm test` fails atm.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -258,6 +258,11 @@ internals.options = function () {
             type: 'boolean',
             description: 'silence test output'
         },
+        'silent-skips': {
+            alias: 'k',
+            type: 'boolean',
+            description: 'don’t output skipped tests'
+        },
         sourcemaps: {
             alias: ['S', 'sourcemaps'],
             type: 'boolean',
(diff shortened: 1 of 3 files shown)


geek:
@rluba well done!

Marsup:
:+1: This is the kind of PR that reminds me we should really consider having a `.labrc`, those CLI options are getting out of hand. (not that there's anything wrong with this PR)

rluba:
@Marsup Yeah, it was not easy to find a free shortcut for the new option — already a pretty long list 😕

PR #552, opened 2016-03-27
Add negative test info to README

apelade:
There are examples in https://github.com/hapijs/catbox-mongodb/blob/master/test/index.js


Marsup:
Why should it belong to lab's docs ? Lab is just the runner, not the assertion framework.
If it's merged, done is in the wrong place in the async example.

apelade:
Agreed, If anything, there could be a negative test in experiment.

PR #556, opened 2016-03-31
Code coverage: Always surround expressions in ternary operator with parenthesis

### LINKED ISSUE #555: Code coverage for conditional operator with comma operator isn't handled correctly
__Version:__ Lab 10.3.0 with code coverage turned on, `-c`

### Problem
This expression is not correctly annotated so when the code is run it produces invalid result

```js
true ? ('no', 'yes') : 'nope'
```

__Expected value:__ 'yes'
__Actual value:__ 'no'

### Why
The reason is it's being annotated like this:

``` js
(global.__$$labCov._statement('/Users/2921/code/_3rd/lab/test/coverage/conditional2.js',1,1,true)
  ? global.__$$labCov._statement('/Users/2921/code/_3rd/lab/test/coverage/conditional2.js',2,1,'no', 'yes')
  : global.__$$labCov._statement('/Users/2921/code/_3rd/lab/test/coverage/conditional2.js',3,1,'nope'))
```

This excerpt shows that _consequent_ expression is missing it's parentheses:

``` js
global.__$$labCov._statement('file.js',2,1,'no', 'yes')
```

It should have been annotated as:

``` js
global.__$$labCov._statement('file.js',2,1,('no', 'yes'))
```

The _consequent_ code is rendered using `node.consequent.source()` on https://github.com/hapijs/lab/blob/b122ae64f2a257d118834cb7a73276b089a8377a/lib/coverage.js#L203

``` js
node.set('(' + node.test.source() + '? global.__$$labCov._statement(\'' + filename + '\',' + consequent + ',' + line + ',' + node.consequent.source() + ') : global.__$$labCov._statement(\'' + filename + '\',' + alternate + ',' + line + ',' + node.alternate.source() + '))');
```

The problem with missing parentheses comes from [acorn](https://github.com/ternjs/acorn) (used by [espree](https://github.com/eslint/espree)) that sets start and end position for _SequenceExpression_ to not include parentheses.

### Solution
I see three possible solutions:

1. Test if _consequent_ node is a _SequenceExpression_. If it is, add parentheses around it. Do the same for _alternate_ node.
2. Always surround _consequent_ and _alternate_ nodes with parentheses, i.e:

   ``` js
node.set('(' + node.test.source() + '? global.__$$labCov._statement(\'' + filename + '\',' + consequent + ',' + line + ',(' + node.consequent.source() + ')) : global.__$$labCov._statement(\'' + filename + '\',' + alternate + ',' + line + ',(' + node.alternate.source() + ')))');
```

3. [acorn](https://github.com/ternjs/acorn) has an option `preserveParens` that could be switched on
  >If this option is `true`, parenthesized expressions
  are represented by (non-standard) `ParenthesizedExpression` nodes
  that have a single `expression` property containing the expression
  inside parentheses."
  
I think the easiest and safest is alternative 2. There shouldn't be any problems with adding extra parentheses around the _consequent_ and _alternate_ nodes. It's already being done around the entire ternary expression.

If you're ok with alternative 2 I can submit a PR with the fix and a test.

### PR DESCRIPTION
**Pull request by HCanber**
Fixes #555 __Code coverage for conditional operator with comma operator isn't handled correctly__

### DIFF
--- a/test/coverage.js
+++ b/test/coverage.js
@@ -175,6 +175,14 @@ describe('Coverage', () => {
         done();
     });
 
+    it('retains original value of conditional result with comma operator', (done) => {
+
+        const Test = require('./coverage/conditional2');
+        const value = 4711;
+        expect(Test.method(value)).to.equal(value);
+        done();
+    });
+
     it('should not change use strict instructions', (done) => {
 
         const Test = require('./coverage/use-strict.js');
--- a/test/coverage/conditional2.js
+++ b/test/coverage/conditional2.js
@@ -1,0 +1,13 @@
+'use strict';
+
+// Load modules
+
+
+// Declare internals
+
+const internals = {};
+
+
+exports.method = function (value) {
+    return true ? (value-1, value) : value+1;  //Should return value
+};
--- a/test/coverage/conditional2.js
+++ b/test/coverage/conditional2.js
@@ -9,5 +9,5 @@ const internals = {};
 
 
 exports.method = function (value) {
-    return true ? (value-1, value) : value+1;  //Should return value
+    return true ? (value - 1, value) : value + 1;  //Should return value
 };
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -200,7 +200,7 @@ internals.instrument = function (filename) {
             consequent = addStatement(line, node.consequent, false);
             const alternate = addStatement(line, node.alternate, false);
 
-            node.set('(' + node.test.source() + '? global.__$$labCov._statement(\'' + filename + '\',' + consequent + ',' + line + ',' + node.consequent.source() + ') : global.__$$labCov._statement(\'' + filename + '\',' + alternate + ',' + line + ',' + node.alternate.source() + '))');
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: test/coverage/conditional2.js [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+// Load modules
+
+
+// Declare internals
+
+const internals = {};
+
+
+exports.method = function (value) {
+    return true ? (value-1, value) : value+1;  //Should return value
```
geek: please space the operators

```js
return true ? (value - 1, value) : value + 1;
```

test/coverage.js
```diff
@@ -175,6 +175,14 @@ describe('Coverage', () => {
         done();
     });
 
+    it('retains original value of conditional result with comma operator', (done) => {
```
geek: You will want to enable coverage on a test that uses the condition2 file to make sure that coverage works and the method executes correctly

HCanber: Not sure what this means. I copied the test above, and just modified it. What do I need to do differently? 
Note, the purpose of this PR is not to add new or missing functionality. If coverage of conditional operator and comma operator was wrong before, it will continue to be wrong after this. This PR just to fix a bug which causes errors (in my case tests starts to fail) when lab is run with coverage switched on.

test/coverage/conditional2.js
HCanber: New commit pushed which fixes this.

geek:
@HCanber thanks for fixing this issue, its now published in lab 10.3.1

PR #573, opened 2016-05-02
use stable stringify for better objects diff

iamdoron:
afterstable

instead of:
beforestable

### DIFF
--- a/lib/reporters/console.js
+++ b/lib/reporters/console.js
@@ -14,11 +14,10 @@ const internals = {
     width: 50
 };
 
-internals.stringify = (obj, serializer, indent, decycler) => {
+internals.stringify = (obj, serializer, indent) => {
 
-    return StableStringify(
-      JSON.parse(StringifySafe(obj, serializer, 0, decycler)),
-      { space: indent });
+    return StableStringify(obj,
+      { space: indent, replacer: StringifySafe.getSerialize(serializer) });
 };
 
 exports = module.exports = internals.Reporter = function (options) {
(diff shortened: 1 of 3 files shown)


Marsup:
This won't work with a circular object.

iamdoron:
what do you mean by 'it won't work'? It should act the same as before

Marsup:
My bad, I thought json-stringify-safe was more clever than it really is.

iamdoron:
Yes, it just outputs `"[Circular]"`.

lib/reporters/console.js
```diff
@@ -13,6 +14,12 @@ const internals = {
     width: 50
 };
 
+internals.stringify = (obj, serializer, indent, decycler) => {
+
+    return StableStringify(
```
Marsup: Can't we just do `StableStringify(obj, { space: indent, replacer: StringifySafe.getSerialize(serializer) })` ?

iamdoron: if you do something like
```javascript
internals.stringify = (obj, serializer, indent) => {

    return StableStringify(
      JSON.parse(StringifySafe(obj)),
      { space: indent, replacer: serializer });
};
```

The test 'console reports with undefined values fails'

After the first stringify we lose some information about the original object:
```javascript
JSON.stringify({a: undefined}) === '{}'
JSON.stringify({a: Infinity}) === '{"a":null}'
```

iamdoron: yes, we can 👍 
I changed it

arb:
I would maybe recommend "fast-safe-stringify" as an alternative since I think it's the fastest one out that that doesn't crash on circular references.

iamdoron:
@arb - using @Marsup  suggestion, it is better to use `StringifySafe.getSerialize()`, as it can be composed with `StableStringify` seamlessly. There's no such interface in `fast-safe-stringify`

geek:
@iamdoron well done, this will be a nice improvement for our console reporter.

PR #584, opened 2016-05-17
Don't execute before/after on experiments that lack tests

### LINKED ISSUE #337: Skipped tests should have their beforeEach() skipped too?
When tests are skipped, their beforeEach() runs anyway. This doesn't make sense to me. Thank you!

### LINKED ISSUE #440: before and after blocks
when i run a specific test using the id
```
./node_modules/lab/bin/lab -v -i 2
```
is it supposed to run all the befores and after blocks??

geek:
Closes #337 
Closes #440

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -464,30 +464,30 @@ internals.executeTests = function (experiment, state, skip, callback) {
 internals.experimentHasTests = function (experiment, state) {
 
     if (experiment.experiments.length) {
-        const experimentsWithTests = experiment.experiments.filter((childExperiment) => {
+        const experimentsHasTests = experiment.experiments.some((childExperiment) => {
 
             return internals.experimentHasTests(childExperiment, state);
         });
 
-        if (experimentsWithTests.length) {
+        if (experimentsHasTests) {
             return true;
         }
     }
 
-    const tests = experiment.tests.filter((test) => {
+    const hasTests = experiment.tests.some((test) => {
 
         if ((state.filters.ids.length && state.filters.ids.indexOf(test.id) === -1) ||
             (state.filters.grep && !state.filters.grep.test(test.title))) {
 
-            return;
+            return false;
         }
 
         if (!test.options.skip && test.fn) {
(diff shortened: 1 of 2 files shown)


lib/runner.js
```diff
@@ -256,7 +256,7 @@ internals.executeExperiments = function (experiments, state, skip, callback) {
         const previousDomains = state.domains;
         state.domains = [];
 
-        const skipExperiment = skip || experiment.options.skip;
+        const skipExperiment = skip || experiment.options.skip || !internals.experirementHasTests(experiment, state);
```
cjihrig: experiment is spelt wrong.

geek: exper ire ment :(

lib/runner.js
```diff
@@ -461,6 +461,29 @@ internals.executeTests = function (experiment, state, skip, callback) {
...
+
+            return;
+        }
+
+        if (!test.options.skip && test.fn) {
+            return test;
+        }
+    });
+
+    return tests.length;
```
cjihrig: You should probably convert this to a Boolean before returning. "has tests" implies to me that it would be a Boolean.

lib/runner.js
```diff
@@ -461,6 +461,36 @@ internals.executeTests = function (experiment, state, skip, callback) {
 };
 
 
+internals.experimentHasTests = function (experiment, state) {
+
+    if (experiment.experiments.length) {
```
AdriVanHoudt: purely personal opinion but I find `> 0` more clear as to what you are doing, just a fyi

lib/runner.js
```diff
@@ -461,6 +461,36 @@ internals.executeTests = function (experiment, state, skip, callback) {
 };
 
 
+internals.experimentHasTests = function (experiment, state) {
```
Marsup: Shall use .some() instead of filters in this, there's no need to have a full count, just need at least one.

geek: good call

PR #591, opened 2016-05-23
Add --lint-fix option.

wprl:
This PR writes any fixes eslint detects to disk if `--lint-fix` is used.

### DIFF
--- a/lib/lint.js
+++ b/lib/lint.js
@@ -3,6 +3,7 @@
 // Load modules
 
 const ChildProcess = require('child_process');
+const Fs = require('fs');
 
 
 // Declare internals
@@ -15,9 +16,19 @@ const internals = {
 exports.lint = function (settings, callback) {
 
     const linterPath = (settings.linter && settings.linter !== 'eslint') ? settings.linter : internals.linter;
-    const child = ChildProcess.fork(linterPath,
-                                  settings['lint-options'] ? [settings['lint-options']] : [],
-                                  { cwd: settings.lintingPath });
+
+    let linterOptions;
+
+    try {
+        linterOptions = JSON.parse(settings['lint-options'] || '{}');
+    }
+    catch (err) {
+        throw new Error('lint-options could not be parsed');
+    }
+
+    linterOptions.fix = settings['lint-fix'];
+
+    const child = ChildProcess.fork(linterPath, [JSON.stringify(linterOptions)], { cwd: settings.lintingPath });
     child.once('message', (message) => {
 
         child.kill();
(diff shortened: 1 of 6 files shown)


wprl:
Tests pass now!

lib/linter/index.js
```diff
@@ -42,17 +42,29 @@ exports.lint = function () {
...
-            filename: result.filePath,
-            errors: result.messages.map((err) => {
+        const transformed = {
+            filename: result.filePath
+        };
+
+        if (result.hasOwnProperty('output')) {
+
+            transformed.fix = {
+                filePath: result.filePath,
```
geek: shouldn't need `filePath` set here, can use the `filename` property of `transformed`

lib/lint.js
```diff
@@ -29,20 +42,27 @@ exports.lint = function (settings, callback) {
...
+                });
+            }
 
             lint.totalErrors = errors;
             lint.totalWarnings = warnings;
             result.totalErrors += errors;
             result.totalWarnings += warnings;
+
+            if (lint.fix) {
+                Fs.writeFileSync(lint.fix.filePath, lint.fix.output);
```
geek: can use `lint.filename`

lib/lint.js
```diff
@@ -14,10 +15,22 @@ const internals = {
 
 exports.lint = function (settings, callback) {
 
+    const linterOptions = settings['lint-options'] ? [settings['lint-options']] : ['{}'];
```
geek: Can this be changed to something like

```js
let linterOptions = JSON.parse(settings['lint-options'] || '{}');
linterOptions = Object.assign(linterOptions, { fix: !!settings['lint-fix'] });
const linterPath = (settings.linter && settings.linter !== 'eslint') ? settings.linter : internals.linter;

const child = ChildProcess.fork(linterPath, [JSON.stringify(linterOptions)], { cwd: settings.lintingPath });
```

geek:
@wprl you will need to update the cli definition with the new flag https://github.com/hapijs/lab/blob/71b927b57258a265107a62ba46a28053fb308785/lib/cli.js#L134  Please also indicate in the readme that the default is to not fix files.

wprl:
How's this look?

lib/lint.js
```diff
@@ -15,9 +16,21 @@ const internals = {
...
-                                  settings['lint-options'] ? [settings['lint-options']] : [],
-                                  { cwd: settings.lintingPath });
+
+    let linterOptions;
+
+    try {
+        linterOptions = JSON.parse(settings['lint-options']);
+    }
+    catch (err) {
+        linterOptions = {};
```
geek: In this case we should throw with an error informing the user that `lint-options` couldn't be parsed.

lib/lint.js
```diff
@@ -15,9 +16,21 @@ const internals = {
...
+    }
+
+    if (settings['lint-fix']) {
+        linterOptions.fix = true;
```
geek: can assign directly now that lint-fix is always a boolean:

```js
linterOptions.fix = settings['lint-fix'];
```

wprl:
OK see how you like this revision :)

PR #593, opened 2016-05-24
Support ability to add notes to a test

### LINKED ISSUE #402: Add ability to attach notes to test results
There are times when a test passes (or fails) but you want to draw attention to some other data about it. My specific need is to record performance numbers (my tests confirm performance is under n milliseconds but I also want to see the actual numbers).

In order to keep the interface and implementation simple, we can just piggyback the `done` argument:

```js
it('runs faster than 500ms', function (done) {
    expect(elapsed).to.be.below(500);
    done.note('Took: ' + elapsed + 'ms');
    done();
});
```

If we make `note()` return `done` we can also chain:
```js
it('runs faster than 500ms', function (done) {
    expect(elapsed).to.be.below(500);
    done.note('Took: ' + elapsed + 'ms')();
});
```

geek:
Closes #402

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,24 @@ lab.experiment('with only', () => {
 });
 ```
 
+The `test()` callback has a `note` function attached to it that can be used to
+attach notes to the test case.  These notes are included in the consoler reporter
+at the end of the output.  For example, if you would like to add a note with the
+current time, your test case may look like the following:
+
+```javascript
+lab.test('attaches notes', (done) => {
+
+    Code.expect(1 + 1).to.equal(2);
+    done.note(`The current time is ${Date.now()}`);
+    done();
+});
+```
+
+Multiple notes can be appended for the same test case by simply calling `note`
+repeatedly.
+
+
 The `test()` callback provides a second `onCleanup` argument which is a function used to register a runtime cleanup function
 to be executed after the test completed. The cleanup function will execute even in the event of a timeout. Note that the cleanup
 function will be executed as-is without any timers and if it fails to call it's `next` argument, the runner will freeze.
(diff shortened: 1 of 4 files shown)


lib/runner.js
```diff
@@ -628,10 +628,18 @@ internals.protect = function (item, state, callback) {
...
+        const done = (err) => {
 
             finish(err, 'done');
-        }, onCleanup);
+        };
+
+        item.notes = [];
+        done.note = (note) => {
+
+            item.notes.push(note);
```
AdriVanHoudt: why not return done as suggested in the issue? seems like a handy thing

Marsup:
Too bad people using promises won't be able to use that.

README.md
```diff
@@ -183,6 +183,24 @@ lab.experiment('with only', () => {
 });
 ```
 
+The `test()` callback has a `note` function attached to it that can be used to
```
hueniverse: `note()`

README.md
```diff
@@ -183,6 +183,24 @@ lab.experiment('with only', () => {
...
+```javascript
+lab.test('attaches notes', (done) => {
+
+    Code.expect(1 + 1).to.equal(2);
+    done.note(`The current time is ${Date.now()}`);
+    done();
+});
+```
+
+Multiple notes can be appended for the same test case by simply calling `note`
```
hueniverse: `note()`

lib/runner.js
geek: for stringing notes together?

`done().notes().notes().notes();` ?

AdriVanHoudt: also but also so you can do `done.note('Took: ' + elapsed + 'ms')();`

Marsup: Agreed, this doesn't hurt having it.

PR #601, opened 2016-06-08
Support .labrc.js for configuration

### LINKED ISSUE #598: Support .labrc files for configuration defaults

geek:
Closes #598


test/run_cli.js
```diff
@@ -11,7 +11,7 @@ const internals = {
 
 module.exports = (args, callback, root) => {
 
-    const cli = ChildProcess.spawn('node', [].concat(internals.labPath, args), { 'cwd' : root ? root : '.' });
+    const cli = ChildProcess.spawn('node', [].concat(internals.labPath, args), { cwd : root ? root : '.' });
```
cjihrig: While you're here... `root ? root : '.'` could just be `root || '.'` I think.

test/cli_labrc/.labrc
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
cjihrig: Does the find rc module support file extensions? Being able to specify .js would be nice at least for syntax highlighting.

README.md
```diff
@@ -367,6 +367,26 @@ if (typeof value === 'symbol') {
 
 ```
 
+## `.labrc` file
+
+**lab** supports a `.labrc` configuration file for centralizing lab settings.  
+The `.labrc` file can be located in the current working directory or in the
+users home directory.  The `.labrc` file needs to be able to be required by
```
cjihrig: user -> user's

cjihrig: The second sentence might start better as "The .labrc file must be requirable by"

cjihrig:
Maybe I'm missing something, but where do I look to see what options are supported by this new file? Are the config options just the name of the long command line flags without the `--` in front?

test/cli_labrc/.labrc
gergoerdosi: I agree, would be good to use an extension. Eslint deprecated `.eslintrc` too in favor of `.eslintrc.json`, `.eslintrc.yaml `, etc.

geek:
I will update the readme to make it clearer what the allowed options are.  They are the same long names as the cli options.

Marsup:
Do you think we should be able to provide our own labrc or do you prefer a base labrc overridden through cli ?

geek:
@cjihrig should be ready to go.  I would rather the file be `.labrc` instead of supporting `.labrc.json` and `.lanrc.js`.  

@Marsup you can provide your own labrc, lab doesn't ship with one.

Marsup:
@geek I didn't mean that. You always have multiple test configuration, how do you intend to deal with that + rc ?

geek:
@Marsup the first one found closest to the heart of the project wins.

1. Current directory
2. Parent of current directory, until the root folder is encountered
3. $HOME/.${apprc}

cjihrig:
LGTM

yelworc:
Quick note: The removed default for `NODE_ENV` caught me by surprise – my tests were suddenly running with the wrong app configuration after updating lab...

geek:
@yelworc thanks for the report.  I will investigate the issue, it shouldn't have happened because there is a default environment set in the runner https://github.com/hapijs/lab/blob/master/lib/runner.js#L40

Issue #606, opened 2016-06-14
NODE_ENV is no longer defaulting to 'test'


nlf:
[this commit](https://github.com/hapijs/lab/commit/4d228729a352a532be29c0a268c21a854610a53d#diff-f8849ae9f0614ba9464fdd14c2d98dc0L180) removed the default for the environment setting erroneously causing NODE_ENV to be set to `undefined` if none is passed

nlf:
the bigger concern here is that there's a test to verify the default that's passing despite this

geek:
@nlf the test is correct, there is a default environment value set in the runner.  It looks like issue is in the merging of the cli config and these runner defaults.

nlf:
ah ok, that makes more sense

geek:
@nlf I actually can't reproduce.  Can you provide more information: the node version and if you have a `.labrc.js` set?  I tried with hapi and it shows that the NODE_ENV is set to "test" when I run using the latest version of lab.

nlf:
using latest lab, node 6.2.1, no `.labrc.js` exists

if you create an empty directory, install lab, and insert the boilerplate below as `test/index.js` it reproduces it for me

```
'use strict';

const lab = exports.lab = require('lab').script();
console.log(process.env.NODE_ENV);
```

geek:
@nlf this is now published

PR #607, opened 2016-06-14
Test for default env

geek:
Relates to #606


Marsup:
@geek this won't demonstrate the bug, spawn is transferring parent process env, since you're running tests in tests you're already set ;)

nlf:
tests are hard

PR #608, opened 2016-06-14
Fix NODE_ENV not being set (fixes #606)

### LINKED ISSUE #606: NODE_ENV is no longer defaulting to 'test'

Marsup:
Naive fix but it works...


geek:
@Marsup thanks, looks good

Issue #459, opened 2015-10-01
`/* `$lab:coverage:(off|on)$` */` disables coverage for a whole module


eriktrom:
Howdy folks - when using the following:

```js
/* $lab:coverage:off$ */
import Path from 'path'
import Something from './somewhere'
/* $lab:coverage:on$ */


// do lots of things that currently have poor test coverage
```

I get 100% coverage for the entire module(in both console and html output). When I remove the `$lab:coverage:(off|on)$` comments, I get what I expect. 

**This seems to indicate that `/* $lab:coverage:on$ */` does not turn coverage back on for the remaining lines.**

(Ancillary note: `import Blah from 'blah'` causes the babel compiled output to fail coverage for the ternary statement it does for imports , that's a separate issue(but it does exist, fyi). Putting off/on comments around the imports seemed like a way forward for now, but then I hit this wall.)

I am using the following tools/versions:

node/v4.1.1
babel/5.8.23 (babel-core 5.8.25)
lab/6.1.0

Note: issue also occurred on lab 5.x.x, I just upgraded to 6.x before making this issue(w/ same result)
Note: I have not tried on older versions of node

My babel transform function (is mostly copied from the README with some slight alterations):

```js
let Babel = require('babel-core');

module.exports = [{
  ext: '.js',
  transform: function(content, filename) {

    let isNodeModules = filename.split('/').indexOf('node_modules') === 0;
    let isTest = filename.split('/').indexOf('test') === 0;
    let isLib = filename.split('/').indexOf('lib') === 0;

    // Make sure to only transform your code or the dependencies you want
    if (isTest || isLib) {
      let result = Babel.transform(content, {
        sourceMap: 'inline',
        filename: filename,
        sourceFileName: filename
      });
      return result.code;
    } else if (isNodeModules) {
      return content;
    }

    throw Error('babel transform has had some issues, content %o, filename %s', content, filename);
  }
}];
```

The lab cli command I run is:

```
./node_modules/lab/bin/lab --sourcemaps --debug --transform ./test/babel-transform.js -a code -t 100 -Lv -r console -o stdout -r html -o coverage.html ./test/unit
```

Word of warning: Babel and Lab are both new toolz I am playing with, forgive me if there is user fault involved, although I tried to make sure this was actually an issue before submitting one.

Let me know if more information is needed. Tell me where to look and I'll help fix it :)

Thanks and cheers!

Marsup:
Can we see the transpiled code ?

[...]

PR #620, opened 2016-08-13
Ensure we're not bypassing loc that should be covered

jedireza:
I experienced a strange case of unexpected 100% test coverage when transpiling (seen at https://github.com/nlf/lab-babel/issues/9). This may also be relevant to [hapijs/lab#459](https://github.com/hapijs/lab/issues/459).

I found that if we place `/* $lab:coverage:(off|on)$ */` in certain positions in the tree, we could accidentally bypass loc that should be covered. This happens because we only check if the node's starting range is in the `bypass` object without ensuring the entire node should be skipped.

By adding an additional check that the node's end range is also included in the `bypass` object, we allow `annotate(...)` to continue recursively processing.

My new test case doesn't use transpilation, but simply puts `/* $lab:coverage:(off|on)$ */` in positions that reproduce the symptom as I found it.

Feedback is much appreciated.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -109,7 +109,7 @@ internals.instrument = function (filename) {
 
         // Coverage status
 
-        if (bypass[node.range[0]]) {
+        if (bypass[node.range[0]] && bypass[node.range[1]]) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


geek:
@jedireza nice find, sorry you ran into this bug :/

PR #648, opened 2016-10-04
Original files in HTML reporter

### LINKED ISSUE #614: Test coverage shows untransformed output
`lab -T test/lib/transform-tsc.js test -v -S -c -r html -o cov.html` produces a coverage file with:
- Correct test results
- Correct pre-transform filenames, e.g. `index.ts` in the navigation bar and code coverage report headers
- Correct coverage information, e.g. reporting "condition always false"
- The post-transform file contents, _not_ the expected pre-transform file contents

`lab -T test/lib/transform-tsc.js test -v -S` with a failing test produces correct line numbers for pre-transformed code.

Reproduction repository: [`garthk/lab-coverage-vs-sourcemaps-typescript-edition`](https://github.com/garthk/lab-coverage-vs-sourcemaps-typescript-edition)

See also: hapijs/lab#338

I'm hoping this is a configuration problem, in which case I'm happy to have pitched in with typings for `code` and `lab` and a skeleton project for anyone trying to drive them from TypeScript and Visual Studio Code.

feugy:
This is a first attempt to fix #614.

My idea is to adapt the HTML report to display original files when source maps is present.
It works well, notably to display uncovered lines or statement.

Sloc number, coverage percent and "covered/not covered" counter are wrong for now, because they are computed from generated files, and not from original files.

I had to update the `while.js` test fixture using latest traceur to make sure we display original source with comment and white spaces, and this explains the small fixes on coverage tests.


test/reporters.js
```diff
@@ -1318,7 +1318,7 @@ describe('Reporter', () => {
 
                 expect(err).not.to.exist();
                 const result = output.replace(/at.*\.js\:\d+\:\d+\)?/g, 'at <trace>');
-                expect(result).to.match(/^\n  \n  x\n\nFailed tests:\n\n  1\) test works:\n\n      actual expected\n\n      {\n\s+"a": 1,\n\s+"b": "\[undefined\]",\n\s+"c": "\[function \(\) \{\\n\\n\s+return 'foo';\\n\s+\}\]",\n\s+"d": "\[Infinity\]",\n\s+"e": "\[-Infinity\]"\n\s+}\n\n      Fail\n\n(?:      at <trace>\n)+\n\n1 of 1 tests failed\nTest duration: \d+ ms\nNo global variable leaks detected\n\n$/);
+                expect(result).to.match(/^\n  \n  x\n\nFailed tests:\n\n  1\) test works:\n\n      actual expected\n\n      {\n\s+"a": 1,\n\s+"b": "\[undefined\]",\n\s+"c": "\[function \(\) \{((\\r)?\\n){2}\s+return 'foo';(\\r)?\\n\s+\}\]",\n\s+"d": "\[Infinity\]",\n\s+"e": "\[-Infinity\]"\n\s+}\n\n      Fail\n\n(?:      at <trace>\n)+\n\n1 of 1 tests failed\nTest duration: \d+ ms\nNo global variable leaks detected\n\n$/);
```
feugy: I'm on Windows, so I had to adapt that huge regexp a little...

test/coverage/while.js
```diff
@@ -10,7 +10,7 @@ const internals = {};
 
 exports.method = function (value) {
 
-    while (value) {
+    while ( value ) {
```
feugy: To make a difference between generated and original file, especially for chunks handling.

feugy:
Sorry, for missing coverage, I'll fix it tomorrow.

As I'm on windows, even when using [ubuntu bash](https://insights.ubuntu.com/2016/03/30/ubuntu-on-windows-the-ubuntu-userspace-for-windows-developers/) I can't run all the tests locally.

It hangs, and sometimes even crash with a good old segfault 😀

geek:
@feugy looks like a great improvement to our reporting. Thanks!

lib/coverage.js
```diff
@@ -412,6 +412,21 @@ internals.addSourceMapsInformation = function (ret, num) {
...
+
+    if (source.chunks) {
+        source.chunks.forEach((chunk) => {
+            // Also add source map information on chunks
+            originalPosition = SourceMapSupport.mapSourcePosition({
+                source: ret.filename,
+                line: num,
+                column: chunk.column
+            });
+            chunk.originalFilename = originalPosition.source.replace(Path.join(process.cwd(), '/').replace(/\\/g, '/'), '');
```
feugy: I've copy-pasted this from [line 404](https://github.com/hapijs/lab/pull/648/files/b95f9616e312574bb1071e8f1337a3ef2a8d0f11#diff-1c3512bd280f1fe9c07481535be0d978R404), but I'll give a try, because using basename is definitely more understandable !

feugy:
Just to let you know: I've found that inlined and external source maps are not reported the same due to my latest modification.

So my TODO list is:

- [ ] support external sourcemaps
- [ ] support inlined sourcemaps
- [ ] fix sloc and coverage counters
- [ ] write tests to maintain the100% coverage

feugy:
Supporting inline maps involved a huge refactor, but it's nearly done.
@geek while trying to figure out how to get sloc and hits in original files, I found that [any lines](https://github.com/hapijs/lab/blob/master/lib/coverage.js#L450) are considered as "line of code", even comments (`data.source` includes the entire source content, not just executable lines).

Did I missed something ?

lib/coverage.js
```diff
@@ -390,18 +391,19 @@ exports.analyze = function (options) {
...
         column: 0
     };
-    const originalPosition = SourceMapSupport.mapSourcePosition(position);
+    let originalPosition = smc.originalPositionFor(position);
     const source = ret.source[num];
 
     if (position !== originalPosition) {
-        source.originalFilename = originalPosition.source.replace(Path.join(process.cwd(), '/').replace(/\\/g, '/'), '');
+        // Ensure folder separator to be url-friendly
+        source.originalFilename = originalPosition.source && originalPosition.source.replace(/\\/g, '/');
```
feugy: Sourcemaps always expect path to use `/` separator, as URLs does, so we need to do the replacement.
Removing process.cwd() is useless as the sourcemap won't contain it.

lib/coverage.js
```diff
@@ -437,10 +453,6 @@ internals.file = function (filename, data, options) {
             source: line
         };
 
-        if (options.sourcemaps) {
-            internals.addSourceMapsInformation(ret, +num);
-        }
```
feugy: Move to the end of the loop, because we need to have the chunk before adding sourcemaps information

lib/reporters/html.js
```diff
@@ -172,7 +277,7 @@ internals.Reporter.prototype.end = function (notebook) {
 internals.findLint = function (lint, file) {
 
     for (let i = 0; i < lint.length; ++i) {
-        if (lint[i].filename.slice(-file.filename.length) === file.filename && lint[i].errors.length) {
+        if (lint[i].filename.replace(/\\/g, '/').slice(-file.filename.length) === file.filename && lint[i].errors.length) {
```
feugy: `lint.filename` is platform dependant, which is ok (notably when we fix lint issues), but file.filename always use `/`, so we need to to the replacement.

lib/reporters/html/partials/cov.html
```diff
@@ -6,6 +6,10 @@
         <div class="hits">{{coverage.cov.hits}}</div>
         <div class="misses">{{coverage.cov.misses}}</div>
     </div>
+    <div id="filters">
+      <input type="checkbox" checked="" onchange="filter(this)" value="generated" id="show-generated">
+      <label for="show-generated">Show generated files</label>
+    </div>
```
feugy: From now, reporter displays both original and generated sources.

This toggle allows to show/hide generated files.

lib/reporters/html/partials/file.html
```diff
@@ -1,5 +1,5 @@
-<div class="file">
-    <h2 id="{{this.filename}}">{{this.filename}}</h2>
+<div class="file {{#if this.sourcemaps}}show generated{{/if}}">
```
feugy: Link to the corresponding generated file.

lib/reporters/html/partials/line.html
```diff
@@ -1,36 +1,11 @@
...
-        <td class="line" data-tooltip>{{@key}}</td>
-        {{> lint this.lintErrors }}
-        <td class="hits" data-tooltip>{{hits this.hits}}</td>
-        <td class="source">{{this.source}}</td>
-        {{#if sourcemaps}}
-            <td class="sourcemaps file" data-tooltip>{{this.originalFilename}}</td>
-            <td class="sourcemaps line" data-tooltip>{{this.originalLine}}</td>
-        {{/if}}
-    </tr>
-{{/if}}
```
feugy: This is a condensed version of previous code, to limit code duplication, although I agree it's less readable.
I've added a unic ID on each line to allow reference from generated files.

One possible improvement is to remove `this.` which are optional (handlebars implicitly look into this)

test/coverage.js
```diff
@@ -91,8 +91,8 @@ describe('Coverage', () => {
         });
 
         expect(missedLines).to.include([
-            { filename: 'test/coverage/while.js', lineNumber: '5', originalLineNumber: 11 },
-            { filename: 'test/coverage/while.js', lineNumber: '6', originalLineNumber: 12 }
+            { filename: 'while.js', lineNumber: '4', originalLineNumber: 13 },
+            { filename: 'while.js', lineNumber: '5', originalLineNumber: 14 }
```
feugy: Due to different test fixture, compiled with babel instead of traceur.

test/coverage.js
```diff
@@ -120,8 +120,7 @@ describe('Coverage', () => {
         });
 
         expect(missedLines).to.include([
-            { filename: './while.js', lineNumber: '5', originalLineNumber: 11 },
-            { filename: './while.js', lineNumber: '6', originalLineNumber: 12 }
+            { filename: 'while.js', lineNumber: '3', originalLineNumber: 8 }
```
feugy: The fact that only one missing line is detected is a possible bug.

I've replaced the `sourcemaps-inline.js` with a newer version processed by babel, and compacted into a single line:
```javascript
'use strict';// Load modules
// Declare internals
const internals={};exports.method=function(value){while(value){value=false;}return value;};
//# sourceMappingURL=data:application/json;base64,...
```
Coverage detects that `while(value)` is always false, but doesn't detect that `value=false` is never reached.

This only happens when generated file is compacted, I'll dig that a little more.

test/coverage.js
```diff
@@ -211,7 +210,7 @@ describe('Coverage', () => {
         const Test = require('./coverage/use-strict.js');
         expect(Test.method.toString()).to.not.contain('13'); // This is the line of the inner use strict
 
-        const testFile = Path.join(__dirname, 'coverage/use-strict.js');
+        const testFile = Path.join(__dirname, 'coverage/use-strict.js').replace(/\\/g, '/');
```
feugy: To run tests on Windows

test/linters.js
```diff
@@ -165,7 +165,7 @@ describe('Linters - eslint', () => {
 
         Fs.writeFileSync = (path, output) => {
 
-            expect(path).to.endWith('test/lint/eslint/fix/success.js');
+            expect(path).to.endWith(Path.join('test', 'lint', 'eslint', 'fix', 'success.js'));
```
feugy: To run tests on Windows

test/reporters.js
```diff
@@ -736,7 +736,7 @@ describe('Reporter', () => {
             Lab.report(script, { reporter: 'console', progress: 2, output: false }, (err, code, output) => {
```
feugy: Both modifications are needed to run tests on Windows

package.json
```diff
@@ -37,10 +38,10 @@
     "lab": "./bin/lab"
   },
   "scripts": {
-    "test": "./bin/_lab -fL -t 100 -m 3000",
+    "test": "node bin/_lab -fL -t 100 -m 3000",
     "posttest": "npm run lint-md",
-    "test-cov-html": "./bin/_lab -fL -r html -m 3000 -o coverage.html",
-    "lint": "./bin/_lab -d -f -L",
+    "test-cov-html": "node bin/_lab -fL -r html -m 3000 -o coverage.html",
+    "lint": "node bin/_lab -d -f -L",
```
feugy: Unfortunately, Windows' cmd and Powershell doesn't understand the `./` syntax.
By using `node` invokation, we ensure a portable command.

test/leaks.js
```diff
@@ -19,25 +27,48 @@ const internals = {
...
+    afterEach((done) => {
+
+        testedKeys.forEach((key) => {
+            // Only delete globals that were manually set, and avoid deleting pre-existing globals
+            if (global[testedKeys] && global[testedKeys] === 1) {
+                delete global[testedKeys];
+            }
+        });
+        done();
+    });
```
feugy: On Windows, `global` keys can actually be deleted, which makes those test not runnable.
I refactored them to:
- automatically delete globals set to 1, and *not* delete the pre-existing values
- check all counter globals existence and absence.

feugy:
Hello @geek, I would like to have your opinion on the current state, and I have a [question](https://github.com/hapijs/lab/pull/648#issuecomment-252266426)

geek:
@feugy I think you found a bug that we are counting comments towards lines of code :/

geek:
@feugy everything here looks like really good changes. After the fix for the sloc, ping me and I will give it one final review before merging.

Awesome work!

corbinu:
@feugy Are you still working on this? If not would be interested in taking over. Thanks!

feugy:
Hi @corbinu 
Yes, but I've very few time these days.

If you want, you can contribute on that PR. Two things are needed: review and [sloc fix](https://github.com/hapijs/lab/pull/648#issuecomment-252182803)

corbinu:
@feugy No worries will try my best and just PR to your repo with my work. In the process of moving to TypeScript so ran into this. Just didn't want to step on your toes ;) so figured I would check first.

feugy:
No offences at all !
In fact, trying this branch against actual Typescript code would be very handy.
I've tried ES6/Babel code so fat, so you might found uncovered edge cases.

geek:
@feugy any luck finding time to look at this PR again... it looks very close to being complete.

feugy:
Hello @geek  !
It's been a long time since my last contribution, all due to a lot corporate work 😄

For Html reporter, I've finally decided to hide sloc counter / coverage percentage on original files, to only show them on generated files. In fact, sourcemaps allow to reverse the original file content, but can't be blindly used to recompute SLOC & coverage.

But this leads me to another (and way more important) fix.
Up to now, lab was considering all lines when computing SLOC indicator, which was easier but not aligned with common definitions [[1]](https://en.wikipedia.org/wiki/Source_lines_of_code#Measurement_methods), [[2]](http://csse.usc.edu/TECHRPTS/2007/usc-csse-2007-737/usc-csse-2007-737.pdf):
> The common definition of physical SLOC describes them as the lines that do not contain blanks or 
comments

I've fixed this, to rely only on the AST output. Now SLOC is:
> Total number of generated lines - blank lines - commented lines
Without considering actual coverage (hits and misses).

All this works (I've added another test), but has an unexpected impact: it mechanically lower the overall coverage percentage.
This may be considered as a breaking change, because some files that were above coverage threshold might be bellow it now that we don't consider comments in the total.

I leave you the final appreciation, I can easily revert the SLOC computation to what it was before.

test/cli.js
```diff
@@ -500,7 +500,7 @@ describe('CLI', () => {
             expect(result.errorOutput).to.equal('');
             expect(result.code).to.equal(1);
             expect(result.output).to.contain('1 tests complete');
-            expect(result.output).to.contain('Coverage: 92.86% (1/14)');
+            expect(result.output).to.contain('Coverage: 90.00% (1/10)');
```
feugy: Due to the new SLOC algorithm

test/coverage.js
```diff
@@ -52,10 +62,10 @@ describe('Coverage', () => {
         Test.method(1, 2, 3);
 
         const cov = Lab.coverage.analyze({ coveragePath: Path.join(__dirname, 'coverage/partial') });
-        expect(Math.floor(cov.percent)).to.equal(63);
-        expect(cov.sloc).to.equal(52);
+        expect(Math.floor(cov.percent)).to.equal(61);
+        expect(cov.sloc).to.equal(49);
         expect(cov.misses).to.equal(19);
-        expect(cov.hits).to.equal(33);
+        expect(cov.hits).to.equal(30);
```
feugy: Due to the new SLOC algorithm

test/coverage.js
```diff
@@ -134,9 +160,9 @@ describe('Coverage', () => {
 
         const cov = Lab.coverage.analyze({ coveragePath: Path.join(__dirname, 'coverage/bypass') });
         expect(Math.floor(cov.percent)).to.equal(100);
-        expect(cov.sloc).to.equal(16);
+        expect(cov.sloc).to.equal(12);
         expect(cov.misses).to.equal(0);
-        expect(cov.hits).to.equal(16);
+        expect(cov.hits).to.equal(12);
```
feugy: Due to the new SLOC algorithm

test/coverage.js
```diff
@@ -146,10 +172,10 @@ describe('Coverage', () => {
         Test.method(1);
 
         const cov = Lab.coverage.analyze({ coveragePath: Path.join(__dirname, 'coverage/bypass-misses') });
-        expect(Math.floor(cov.percent)).to.equal(93);
-        expect(cov.sloc).to.equal(16);
+        expect(Math.floor(cov.percent)).to.equal(76);
+        expect(cov.sloc).to.equal(13);
         expect(cov.misses).to.equal(1);
-        expect(cov.hits).to.equal(15);
+        expect(cov.hits).to.equal(10);
```
feugy: Due to the new SLOC algorithm

test/reporters.js
```diff
@@ -1477,13 +1477,13 @@ describe('Reporter', () => {
 
                 expect(err).not.to.exist();
                 expect(output).to.contain('<div class="stats medium">');
-                expect(output).to.contain('<span class="cov medium">71.43</span>');
+                expect(output).to.contain('<span class="cov medium">66.67</span>');
```
feugy: Due to the new SLOC algorithm

geek:
@feugy looks good so far, I am going to take a closer look later and merge/release.

geek:
@feugy do we need the `report-*.html` files in the root?

geek:
@feugy if you run the html report on lab it will report `103.31%` coverage for `lib/index.js`

feugy:
@geek about the report.html, its a mistake (test outputs I've forgot to delete). About the coverage, I'm more concerned: there must be an some lines I've considered as commented and that were hit.

feugy:
@geek, nicely spotted !
The bug was pretty obvious: blank lines within comments were subtracted twice: one as commented line, one as blank lines.

It was the opportunity for me to refactor the code, and use only the AST to detect commented lines.
This solution is simpler and more reliable, because detecting comments during coverage pass is tricky.

Can you please give a try ?

Issue #652, opened 2016-10-21
disable lab before timeout


danielo515:
According to the documentation, lab.before accepts an option object with arguments, one of them can be timeout

> timeout - set a specific timeout in milliseconds. Disabled by default or the value of -M.

According to documentation it is disabled by default, but if I do not set certain timeout it is failing after 2k milliseconds. I tried to set it to 0 with the same result. Could someone teach me a valid way?

Thanks and regards

geek:
It looks like a bug where 0 isn't allowed: https://github.com/hapijs/lab/blob/master/lib/runner.js#L338

I will fix this later tonight, or if you want to submit a PR, I am happy to merge and publish.

kgraves:
I'm happy to submit a PR for this, if you guys haven't already done the work. However, I have a couple questions:

* On the above quote of the documentation, I'm confused as to what the `-` in `-M` represents.
* What, if any, constraints do we put on `timeout`? (i.e. positive number, integer, etc)

geek:
@kgraves 
`-M` represents the command argument for the timeout for each test
It needs to be anything that `setTimeout` will accept

danielo515:
I think this issue can be closed

danielo515:
Sorry for issue respawn, but The documentation is still wrong about this feature.

What documentation says.

> Timeout - set a specific timeout in milliseconds. Disabled by default or the value of -M.

But my experience demonstrates that it is **not** Disabled by default, if you do not provide a value, and relies to the default behavior it fails after 2k milliseconds. I suggest to update the documentation or to update the code to match the documentation.

geek:
@danielo515 please open a PR to fix the docs, I'll merge.

PR #653, opened 2016-10-26
fixes logic for lab.before timeout #652

kgraves:
Here's my first stab at fixing the logic. (#652)  I didn't see anything in the Hapi style guide on ternary operators, so I left it on a single line.

It seems to work fine, but for some reason one test is failing on me, even when I have a fresh repo without my changes:

```bash
Failed tests:

  65) Coverage identifies lines with partial coverage when having inline sourcemap:

      Expected [ { filename: 'test/coverage/while.js',
    lineNumber: '5',
    originalLineNumber: 11 },
  { filename: 'test/coverage/while.js',
    lineNumber: '6',
    originalLineNumber: 12 } ] to include [ { filename: './while.js',
    lineNumber: '5',
    originalLineNumber: 11 },
  { filename: './while.js',
    lineNumber: '6',
    originalLineNumber: 12 } ]

      at it (/home/kg/Documents/gits/lab/test/coverage.js:122:32


1 of 296 tests failed
Test duration: 35270 ms
No global variable leaks detected
Coverage: 100.00%
Linting results: No issues
```

I'm assuming it's not caused by my changes, however Travis-ci doesn't seem to be failing...
Any ideas?

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -335,7 +335,7 @@ internals.executeDeps = function (deps, state, callback) {
 
     Items.serial(deps, (dep, next) => {
 
-        dep.options.timeout = dep.options.timeout || state.options['context-timeout'];
+        dep.options.timeout = Hoek.isInteger(dep.options.timeout) ? dep.options.timeout : state.options['context-timeout'];
         internals.protect(dep, state, next);
     }, callback);
 };
--- a/test/runner.js
+++ b/test/runner.js
@@ -1797,5 +1797,31 @@ describe('Runner', () => {
                 done();
             });
         });
+
+        it('setTimeout still functions correctly with non-integer timeout', (done) => {
+
+            const script = Lab.script();
+            script.before(overrideGlobals);
+
+            script.after(resetGlobals);
+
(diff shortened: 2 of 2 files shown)


kgraves:
It looks like it *MAY* have to do with `source-map-support` introducing some issues between 0.4.2 and 0.4.5 (current). [see here](https://github.com/evanw/node-source-map-support/issues/153)

The last build of Lab on Travis-ci used 0.4.2 (see [here](https://travis-ci.org/hapijs/lab/jobs/162856996#L415)), which was about a month ago.

Lab's package.json specifies 0.4.x.
I just ran the tests locally with `source-map-support` pinned at 0.4.2 and all tests pass.


Any ideas how to proceed? should I pin `source-map-support` at 0.4.2?

geek:
@kgraves your change looks good. If you can add a test for the non-integer case then that will help with the coverage. I will address the other issue with travis and 0.4.5.

kgraves:
@geek I've added a unit test. Let me know if there's anything else missing. :beer:

Issue #649, opened 2016-10-19
Make the test fail on a Promise unhandled rejection


DanReyLop:
We've recently migrated from NodeJS `4.x.x` to `6.7.0` and we found that in some cases our Promise-based code was failing silently. In recent versions of NodeJS, when a Promise is rejected but it doesn't have a `.catch()` handler, a warning is emitted in the console. But that warning was the only way to notice it, and it was easy to overlook, because the tests didn't fail.

I propose that `lab` subscribes to the `process.on('unhandledRejection')` event (documented [here](https://nodejs.org/api/process.html#process_event_unhandledrejection), available since NodeJS `1.4.1`) and make the test fail if any Promise was rejected and not handled.

Even though what I propose is a breaking change, I think it's worth going that way. 99% of time, if your code under test has a Promise being rejected and fail silently, there's something wrong with that code that ought to be fixed.

Any opinions on this? I'll be happy to make a PR, but I want to have an idea beforehand that the solution isn't gonna be rejected completely :)

geek:
@DanReyLop I like the idea, but I don't want to break applications that already have an `unhandledRejection` handler. Instead, it would be nice if the warning was backported in Node.js itself, but that's not likely to happen.

An alternative is for anyone using promises to add
```js
process.on('unhandledRejection', (reason, promise) => {
  console.error(reason);
  throw reason;
});
```

Honestly, my bigger concern is that once we add this change people will start opening issues to lab saying that it broke their promises since we will be throwing when people don't handle rejections.

DanReyLop:
> I like the idea, but I don't want to break applications that already have an unhandledRejection handler.

Good point. I've yet to see any of those applications in the wild, but they may exist :)
We could check, when the `unhandledRejection` event is fired, there is another handler apart from `lab`'s, and don't fail the test in that case. I think this is starting to sound more complex than it should...

> Instead, it would be nice if the warning was backported in Node.js itself, but that's not likely to happen.

The problem with the warning is that it's very easy to miss. In a `lab` run, you normally see a fast scrolling list of checkmarks and, at the end, either it passes or it fails. If the test suite doesn't fail, very few people would scroll back to find some warnings.

> Honestly, my bigger concern is that once we add this change people will start opening issues in lab saying that it broke their promises since we will be throwing when people don't handle rejections.

I understand that this, if implemented, would be a `major` semver change. That and clearly stating that the error is caused by an unhandled promise _may_ be enough.

[...]

PR #658, opened 2016-11-04
Added a flag (-R, --rejections) to make the test fail if an unhandled Promise rejection happened during its execution

DanReyLop:
Fixes https://github.com/hapijs/lab/issues/649

I'm open to suggestions, specially about naming.

All the relevant discussion is on the #649 issue. To sum up, I've added a flag to Lab, called `--rejections` (alias `-R`). If that flag is set, and an unhandled Promise rejection happens during a test, that test will be failed.

Without this flag, the behaviour will be exactly the same as before: The test pass, and in versions of Node.JS `6.6.0+`, a warning is printed on the console (but the test still passes).

An unhandled Promise rejection looks like this:
```
function foo() {
    functionThatReturnsPromise()
        .then((res) =>{
            throw new Error('This error will be absolutely swallowed.');
        })
        // No .catch() block here
}
```
Initially, the behaviour of unhandled rejections was to be silenced and totally ignored, by design. It looks like everyone is changing their minds, because in the latest versions of Chrome and Node.JS (maybe more engines, I don't know more examples) the unhandled rejections have started to display console warnings. In 99% an unhandled rejection means that something is terribly bad with the code, and the fact that doesn't crash the entire process is just a questionable design choice. So I think it should definitely make a test fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,10 +37,10 @@
     "lab": "./bin/lab"
   },
   "scripts": {
-    "test": "./bin/_lab -fL -t 100 -m 3000",
+    "test": "node ./bin/_lab -fL -t 100 -m 3000",
     "posttest": "npm run lint-md",
-    "test-cov-html": "./bin/_lab -fL -r html -m 3000 -o coverage.html",
-    "lint": "./bin/_lab -d -f -L",
+    "test-cov-html": "node ./bin/_lab -fL -r html -m 3000 -o coverage.html",
+    "lint": "node ./bin/_lab -d -f -L",
     "lint-md": "eslint --config hapi --rule 'strict: 0, eol-last: 0' --plugin markdown --ext md  ."
   },
   "license": "BSD-3-Clause"
--- a/test/cli.js
+++ b/test/cli.js
@@ -1024,10 +1024,7 @@ describe('CLI', () => {
 
         RunCli(['test/cli_reject_promise/reject_promise.js'], (error, result) => {
 
-            if (error) {
-                done(error);
-            }
-
+            expect(error).to.not.exist();
             expect(result.code).to.equal(0);
             done();
         });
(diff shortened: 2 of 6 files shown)


DanReyLop:
Ouch. I've added tests to cover this feature in `test/cli.js`, but the coverage isn't counting those tests (I assume because they spawn child processes). I haven't realized it because as a Windows user I couldn't get `npm test` to pass 😄 

Do you have any suggestion on how to handle this? I thought about adding tests to `test/runner.js` directly, but since what's being tested here is a CLI flag, it made more sense to test it in `test/cli.js`

test/cli.js
```diff
@@ -1019,4 +1019,30 @@ describe('CLI', () => {
...
+            expect(result.code).to.equal(0);
+            done();
+        });
+    });
+
+    it('fails with an unhandled Promise rejection if the specified flag is set', (done) => {
+
+        RunCli(['test/cli_reject_promise/reject_promise.js', '-R'], (error, result) => {
+
+            if (error) {
```
geek: instead:

```js
expect(error).to.not.exist();
```

geek:
@DanReyLop The naming looks good. For improving the code coverage, you are likely going to need to emit the rejection event manually on the `process`, and perhaps monkey-patch other areas.

DanReyLop:
@geek I've ended up adding a test to `tests/runner.js` to cover the new code and actually make it count for the coverage. Coverage is at 100% again 😄

package.json
```diff
@@ -37,10 +37,10 @@
     "lab": "./bin/lab"
   },
   "scripts": {
-    "test": "./bin/_lab -fL -t 100 -m 3000",
+    "test": "node ./bin/_lab -fL -t 100 -m 3000",
     "posttest": "npm run lint-md",
-    "test-cov-html": "./bin/_lab -fL -r html -m 3000 -o coverage.html",
-    "lint": "./bin/_lab -d -f -L",
+    "test-cov-html": "node ./bin/_lab -fL -r html -m 3000 -o coverage.html",
```
DanReyLop: Not directly related to this PR, but the previous command didn't work on Windows, because the `#! /bin/node` trick obviously doesn't work on Windows.

Some tests keep failing on my Windows 10 machine, but with this change at least I can run them.

PR #660, opened 2016-11-07
Default seed is more user-friendly

geek:


rmehner:
👍  didn't think of that, good work :)

PR #662, opened 2016-11-11
fix: don't show seed when shuffle was not enabled

rmehner:
Previously the seed was always shown in the output, even when the order
was not randomized. This could be confusing. With this change, the seed
will only be included in the output if shuffle was enabled.


geek:
Good catch

Issue #668, opened 2017-01-15
.labrc.js with --harmony options execute failed


TangMonk:
this following is config file:
```
module.exports = {
    coverage: true,
    threshold: 90,
};
```

when I run `node  --harmony ./node_modules/.bin/lab test/server/api/index.js`

it will throw error: 

```
SyntaxError: Unexpected token =>
    at Parser.instance.raise.instance.raiseRecoverable (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.3.3.2@espree/espree.js:408:19)
    at Parser.instance.unexpected (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.3.3.2@espree/espree.js:441:14)
    at Parser.pp.semicolon (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.4.0.4@acorn/dist/acorn.js:593:59)
    at Parser.pp$1.parseVarStatement (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.4.0.4@acorn/dist/acorn.js:963:8)
    at Parser.pp$1.parseTopLevel (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.4.0.4@acorn/dist/acorn.js:658:23)
    at Object.parse (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.3.3.2@espree/espree.js:735:21)
    at Object.internals.instrument (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.11.2.2@lab/lib/coverage.js:238:25)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.11.2.2@lab/lib/coverage.js:31:55)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Function._load (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/node_modules/.0.6.8@pmx/lib/transaction.js:62:21)
    at require (internal/module.js:20:19)
    at Object.exports.register (/Users/tangmonk/Documents/mygit/nodejs/hotel_didi_car/server/api/index.js:5:122)
[...]

PR #670, opened 2017-01-16
upgrade espree parser to use ECMAScript latest 8

TangMonk:
#668


geek:
Marking this as v12.0.0 as this may change the coverage % for some people.

PR #671, opened 2017-01-17
Add --bail support

### LINKED ISSUE #214: Stop Tests On First Failure
The idea would be to add a command line flag to lab that would cause the set of test cases to stop on first failure and show the output of that failed test instead of having to wait for all tests to complete and then see the output.

You could also make it an option that it shows the reason for the test failure in line with the verbose output so you can see the reason for the failure as the tests continue to run so you can determine if tests are failing due to the same issue. Again without waiting for the entire set of tests to complete.

geek:
Closes #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ global manipulation. Our goal with **lab** is to keep the execution engine as si
 
 **lab** supports the following command line options:
 - `-a`, `--assert` - name of assert library to use.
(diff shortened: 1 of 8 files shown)


geek:
@Marsup thanks for the feedback, the PR is updated.

Marsup:
Looks even simpler 👍

lerouxb:
Great. Now can someone please do a release so I can use this? :)

geek:
@lerouxb published!

PR #674, opened 2017-01-21
Typos and inspect supports port

geek:


test/cli.js
```diff
@@ -303,6 +305,47 @@ describe('CLI', () => {
         });
     });
 
+    it('starts the inspector with --inspect', (done) => {
```
cjihrig: Why not just hit the `/json/list` endpoint?

geek: It can do that, but there isn't a need to add the extra check, the output should be enough.

README.md
```diff
@@ -507,11 +507,11 @@ number of incomplete assertions, their location in your code and return a failur
...
+This debugger can be accessed using the URL that is printed in the console, or used in association with a few Chrome extensions ([Node.js V8 Inspector](https://chrome.google.com/webstore/detail/nodejs-v8-inspector/lfnddfpljnhbneopljflpombpnkfhggl), [NIM](https://chrome.google.com/webstore/detail/nim-node-inspector-manage/gnhhdgbaldcilmgcpfddgdbkhjohddkj/related), etc).
 
-As you may know, if your tests are associated with the command `npm test`, you can already run `npm test -- --inspect` to run it with the inspector and avoid creating another command.
+As you may know, if your tests are associated with the command `npm test`, you can already run `npm test -- --inspect` to run it with the inspector and avoid creating another command. If you want to listen on a specific port for the inspector, pass `--inspect={port}`. If you wish to have the inspector break on the first test script, you can pass `--inspect-brk`. Similarly, if you want to listen on a specific port with breaking on the first script use `--inspect-brk={port}`.
```
cjihrig: `--inspect-brk` isn't available yet is it?

test/cli.js
```diff
@@ -303,6 +305,47 @@ describe('CLI', () => {
...
+
+        const httpServer = new Http.Server(() => {});
+        httpServer.listen(0, () => {
```
cjihrig: This is trying to start the inspector on a random/available port, right? It kinda has a race condition, even though you're unlikely to run into it.

geek: no race, since it only calls the next function when the server stops

cjihrig: Yea. The race condition is with other things on the system that could grab the port.

README.md
geek: updated, take a look

README.md
```diff
@@ -25,13 +25,14 @@ global manipulation. Our goal with **lab** is to keep the execution engine as si
 - `-C`, `--colors` - enables or disables color output. Defaults to console capabilities.
 - `-d`, `--dry` - dry run. Skips all tests. Use with `-v` to generate a test catalog. Defaults to `false`.
 - `-D`, `--debug` - print the stack during a domain error event.
+- `--debug-brk` - break on the first line of application code.
```
cjihrig: Might want to specify that this is not the same as the built in `--debug-brk`. The built in will break on the first instruction of lab, while lab's version will break in the actual test (if I understand this correctly).

test/cli.js
```diff
@@ -303,6 +310,47 @@ describe('CLI', () => {
         });
     });
 
+    it('starts the inspector with --inspect', { skip: (NODE_MAJOR <= 4) }, (done) => {
```
cjihrig: I believe the inspector first appeared in v6.3.0. I think the skip condition should be "good enough" for CI, but could break if someone was running on v6.1.0 for example.

test/cli.js
```diff
@@ -303,6 +310,47 @@ describe('CLI', () => {
...
+        });
+
+        const startInspector = function (port) {
+
+            const labPath = Path.join(__dirname, '/../bin/lab');
+            const testPath = Path.join(__dirname, '/cli_inspect');
+            const childEnv = Object.assign({}, process.env);
+            delete childEnv.NODE_ENV;
+            const cli = ChildProcess.spawn(labPath, [].concat([testPath, `--inspect=${port}`]), { env: childEnv, cwd : '.' });
+            let combinedOutput = '';
```
cjihrig: We don't actually need the combined output do we?

README.md
Marsup: It's passed to node so it is the same. I don't see any code intercepting the 1st test call.

Marsup: Do you really think adding this option is worth the complexity ? I never found myself debugging without this instruction, which is why I included it by default. Nor did I ever want to change port, why do that ?

cjihrig: If it's the same as Node's built in, then we shouldn't document it. I'm not sure it's the behavior we'd want either.

bin/lab
```diff
@@ -2,28 +2,31 @@
...
     if (process.env.NODE_DEBUG_OPTION) { // WebStorm debugger
         args.push.apply(args, process.env.NODE_DEBUG_OPTION.split(' '));
         delete process.env.NODE_DEBUG_OPTION;
     }
-    else { // Node.js native debugger
-        args.push(
-            '--inspect', // node's inspect
-            '--debug-brk' // you probably want to break 1st if you want to have a chance to hook onto the debugger before the tests run
-        );
+    else { // Node.js native inspector & debugger
```
Marsup: "V8 debugger" if that's what it's called.

README.md
geek: This was added for testing purposes. I will try again with the debug-brk auto added, but I couldn't kill the child process since it was hung on a brk

geek: @Marsup this changes things a bit: https://github.com/nodejs/node/blob/master/src/node_debug_options.cc#L95-L107

test/cli.js
geek: For some reason the message is sent to `stderr`

README.md
```diff
@@ -505,13 +505,13 @@ number of incomplete assertions, their location in your code and return a failur
 
 ## Debuggers
 
-**lab** can be started with the option `--inspect` which will run it with the Node.js native debugger enabled, breaking on the 1st instruction.
+**lab** can be started with the option `--inspect` which will run it with the V8 Inspector and Node.js debugger enabled
```
Marsup: Duplicate wording ?

bin/lab
```diff
@@ -2,28 +2,31 @@
...
         args.push.apply(args, process.env.NODE_DEBUG_OPTION.split(' '));
         delete process.env.NODE_DEBUG_OPTION;
     }
-    else { // Node.js native debugger
-        args.push(
-            '--inspect', // node's inspect
-            '--debug-brk' // you probably want to break 1st if you want to have a chance to hook onto the debugger before the tests run
-        );
+    else { // V8 inspector
+        args = args.concat([inspectArg, '--debug-brk']);
```
Marsup: I feel like we can go back to my const + push, can't we ?

PR #688, opened 2017-03-14
Add failing test for conditional value

Marsup:
We have a similar case on our project where a value is assigned based on boolean short-circuits, it shows it as uncovered so we have to convert it to `A ? A : B` instead of `A || B`. I looked at it quickly but I'm not sure how to fix it or if we want to fix it at all, as it could detect some actual failures in coverage. I'm PRing the failing test in case someone has an idea.

### DIFF
--- a/test/coverage.js
+++ b/test/coverage.js
@@ -312,6 +312,21 @@ describe('Coverage', () => {
         done();
     });
 
+    it('should measure coverage on conditional value', (done) => {
+
+        const Test = require('./coverage/conditional-value');
+        expect(Test.method(1)).to.be.undefined();
+        expect(Test.method(0)).to.be.undefined();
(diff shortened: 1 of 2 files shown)


Marsup:
Hmmm, did you really intend to merge that ? :)

geek:
@Marsup ya, extra motivation to resolve the issue... if it doesn't work out I'll revert.

geek:
@Marsup thoughts? https://github.com/hapijs/lab/pull/689/files

Marsup:
I don't get it, you fixed the failing test which demonstrated the issue but not the code.

geek:
@Marsup the test demonstrates that `A || B` is covered. Is the issue only when `B` is a default value that never changes?

Marsup:
Not sure, but that's the only difference I see.

PR #698, opened 2017-04-13
Improve bail strategy not to execute next befores/afters

Marsup:
Not sure if it should be bug or enhancement, I'll let you decide the label.


geek:
Looks good, thanks for the fix

Issue #700, opened 2017-05-01
Promises that return values now error on finished


corbinu:
The latest change in 13.0.3 has caused tests that return a promise that returns an value to cause an error. Looks like it was working from 10.1.0 on.

```
 Non Error object received or caught

      at Immediate.finish (/home/corbinu/project/node_modules/lab/lib/runner.js:646:19)
      at runCallback (timers.js:676:20)
      at tryOnImmediate (timers.js:645:5)
      at processImmediate [as _immediateCallback] (timers.js:617:5)
```

for example a test
```
function getVals() {
    return new Promise((resolve) => {
        resolve([1, 2, 3]);
    });
}

lab.test("Returns value", () => {

    return getVals().then((vals) => {
        expect(vals).to.be.an.array();

        return vals;
    });

});
```

Marsup:
Indeed, my bad. At least now we have a test for it :)

corbinu:
No worries happy to help :)

geek:
@corbinu thanks for the help, this is now released as v13.0.4

jcollum:
This will also happen if you add `done` to a promise chain: 

```js
 Promise.mapSeries( //bluebird 
          [
            ... many promises
          ],
          () => {} //iterator, don't care about iteration here 
        )
          .then(done) // oops, done is called with the results of the mapSeries call!! 
          .catch(err => {
            console.error(err);
            done(err);
          });
      });

 ```

Solution: 

```js
 .then(() => done() )
```

That error is mighty cryptic though. Could it be changed to "done called with non-Error object"?

PR #701, opened 2017-05-01
Ignore while a promise returns inside a test with a value. Only watch for rejection

corbinu:
Addresses #700 by adding a new test for a resolving promise and then updating the runner to ignore resolved values. Let me know if there are any changes you want me to make :)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -743,7 +743,7 @@ internals.protect = function (item, state, callback) {
         if (itemResult &&
             itemResult.then instanceof Function) {
 
-            itemResult.then(done, done);
+            itemResult.then(() => done(), done);
         }
         else if (!item.fn.length) {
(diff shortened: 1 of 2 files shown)


lib/runner.js
```diff
@@ -743,7 +743,11 @@ internals.protect = function (item, state, callback) {
         if (itemResult &&
             itemResult.then instanceof Function) {
 
-            itemResult.then(done, done);
+            itemResult.then(() => {
```
Marsup: Just `itemResult.then(() => done(), done)`.

corbinu:
@Marsup Your suggested change seems to have caused 2 other tests to fail. Thoughts?

Marsup:
Your change doesn't look remotely similar to my suggestion :)

Marsup:
Thanks ! I'll let @geek merge and release as it's not mine to make.

PR #713, opened 2017-05-31
Bundle code as default assertion library

### LINKED ISSUE #712: Include code assertion library
Most users of lab use code, we should include it and export the expect function from lab. Additionally, we should default to code as the assertion library when running tests. People should be able to opt-out of code as the assertion library as they can today. This will be breaking as the default test behavior will change to enable assertion checks by default

geek:
Closes #712

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -13,6 +13,7 @@ const Runner = require('./runner');
 const Transform = require('./transform');
 const Utils = require('./utils');
 // .labrc configuration will be required if it exists
+// index.js required below if setting assertions module
 
 
 // Declare internals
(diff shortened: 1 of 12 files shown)


geek:
@cjihrig mind looking again, I added `--assert false` to support disabling the assertion library from the CLI

lib/cli.js
```diff
@@ -439,7 +440,11 @@ internals.options = function () {
         options.output = process.stdout;
     }
 
-    if (options.assert) {
+    if (options.assert && options.assert.toLowerCase() === 'false') {
```
cjihrig: Maybe combine the `if` and `else if` into a single `if (options.assert)`, then branch inside there.

PR #722, opened 2017-06-23
Expose Code features directly on script

Marsup:
I was frustrated not to be able to do :
```js
const { describe, it, before, after, expect } = exports.lab = require('lab').script();
```
So there's that.


geek:
@Marsup samesies... this is now published.

PR #726, opened 2017-07-12
Fix npm 5.2 error

Marsup:
npm 5.2 stopped hoisting modules or whatever, so eslint doesn't find its modules anymore, I think having the full path satisfies whatever npm version there is.

### DIFF
--- a/lib/linter/.eslintrc.js
+++ b/lib/linter/.eslintrc.js
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


Marsup:
Failure unrelated to this PR.

Marsup:
It's going to be reverted (see https://github.com/npm/npm/issues/17717). I don't know if it's useful to keep that then. Thoughts ?

cjihrig:
It might be more bullet proof to keep this change in place.

PR #740, opened 2017-08-25
add object spread option

### LINKED ISSUE #739: Coverage is broken when using object spread operator
When executing lab with coverage and requiring a file within the test making use of object spread it fails with:

```
Unexpected token ...
```

Seems that there has been a similar bug in the past #514.

wzrdtales:
Maybe it would make sense to let users pass in their own config.

Fixes #739

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -252,7 +252,7 @@ internals.instrument = function (filename) {
         comment: true,
         range: true,
         ecmaFeatures: {
-          experimentalObjectRestSpread: true
+            experimentalObjectRestSpread: true
         },
         ecmaVersion: 8
     });
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -251,6 +251,9 @@ internals.instrument = function (filename) {
         loc: true,
         comment: true,
         range: true,
+        ecmaFeatures: {
+          experimentalObjectRestSpread: true
+        },
         ecmaVersion: 8
     });
 


wzrdtales:
sry forgot to deactive prettier... overwritten now with an unfixed version

lib/coverage.js
```diff
@@ -251,6 +251,9 @@ internals.instrument = function (filename) {
         loc: true,
         comment: true,
         range: true,
+        ecmaFeatures: {
+          experimentalObjectRestSpread: true
```
geek: need to indent this a tad more, that is why the CI is failing

wzrdtales: yeh before prettier died reindint just everything ;) Would be awesome if you could add an .eslintrc to your project so prettier can do that stuff for me :)

wzrdtales:
Now it fails b/c it actually lints /home/travis/build/hapijs/lab/README.md oO

lib/coverage.js
cjihrig: https://github.com/hapijs/lab/blob/master/lib/linter/.eslintrc.js

wzrdtales: @cjihrig That should be in the project root actually.

cjihrig: Nonetheless, that's where it is if you want to access it.

geek:
@wzrdtales published as 14.2.1

wzrdtales:
+1

temsa:
🎉

PR #743, opened 2017-08-25
Document verbosity value

### LINKED ISSUE #737: Please add an explanation of verbosity; what does 'verbosity: 3.00' mean?
I found it in the PR `notebook.assertions / totalTests`, so I'm doing 3 assertions per test on average. 

OK, but what does it mean? Is more verbosity good? Is there a number that's considered "good enough"? I'm just curious because I've never heard this term before. 

[Googling for it just brings up a ton of python pages.](https://www.google.com/search?q=what+is+unit+test+verbosity&oq=what+is+unit+test+verbosity+&aqs=chrome..69i57.5173j0j7&sourceid=chrome&ie=UTF-8)

geek:
Fixes #737

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,6 +536,8 @@ describe('expectation', () => {
 This is an invalid test but it will pass as the `.false` assertion was not actually called. Lab will report the
 number of incomplete assertions, their location in your code and return a failure of the tests.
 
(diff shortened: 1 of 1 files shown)


jcollum-cambia:
Awesome, thanks.

Issue #759, opened 2017-09-25
More helpful message when test path not found


jtlapp:
I'll be following up with a PR

I happened to have a library file called `test.js` and my tests in a directory called `tests/`. When lab spat out the following error, I went investigating `test.js`, particularly because I usually exclude the filename extension from my module `require`s.:

```
fs.js:954
  binding.stat(pathModule._makeLong(path));
          ^

Error: ENOENT: no such file or directory, stat 'test'
    at Object.fs.statSync (fs.js:954:11)
    at traverse (/usr/local/lib/node_modules/lab/lib/cli.js:73:29)
    at paths.forEach (/usr/local/lib/node_modules/lab/lib/cli.js:103:38)
    at Array.forEach (<anonymous>)
    at Object.internals.traverse (/usr/local/lib/node_modules/lab/lib/cli.js:101:11)
    at Object.exports.run (/usr/local/lib/node_modules/lab/lib/cli.js:62:31)
    at Object.<anonymous> (/usr/local/lib/node_modules/lab/bin/lab:50:23)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
    at Function.Module.runMain (module.js:609:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:598:3
```

The problem was that I'd forgotten to specify the test path, as in `lab tests`. I guess I was also inclined to believe that uncaught exceptions would be thrown from my code, representing problems with the code.

The solution to the problem is easy: just have lab report a friendly error message that makes the problem clear. Okay, now to submit the PR.

PR #760, opened 2017-09-25
Helpful message when path not found

jtlapp:
This is a PR for issue #759. Also appended LFs to error messages that `cli.js` outputs to `options.output` so that the shell prompt lands on the following line.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -111,7 +111,7 @@ internals.traverse = function (paths, options) {
             throw err;
         }
         options.output.write('Could not find test file or directory \'' + lastPath + '\'.\n');
-        process.exit(0);
+        process.exit(1);
     }
 
     if (options.pattern && !testFiles.length) {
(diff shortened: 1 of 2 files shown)


jtlapp:
A decision still needs to be made about the appropriate exit code. If it should be zero, then the reporter test case cited in #759 will fail (and apparently for the wrong reasons).

jtlapp:
An alternative solution would pull the missing file name out of the error message, so that lab would properly report files in subdirectories that get deleted in the middle of a test run (under very specific circumstances). I was less comfortable with creating a dependency on the form of the error message. The only other option adds complexity to the `traverse()` function.

The present implementation seemed like a reasonable compromise of probability and simplicity.

jtlapp:
Hold on. I thought of a simple fix that shows even files deleted mid-test. Mod coming...

jtlapp:
That should do it.

geek:
@jtlapp thanks, this will be included in the next release.

PR #766, opened 2017-10-12
Remove domains and callbacks

### LINKED ISSUE #752: Remove domains in favor of try/catch on async/await

### LINKED ISSUE #765: Use of onCleanup() when test returns a promise
Currently the API to register a cleanup routine for a test is tangled with the use of callback-style tests (`done()`).  That's because `onCleanup` is passed as the second argument to each test, and [this check](https://github.com/hapijs/lab/blob/aaf18fe5e3cdee51a5d9a9468127327252637d36/lib/runner.js#L742-L748) forbids the test to both return a promise and receive any arguments.

It would be very useful to be able to register a cleanup routine and use the promise API, especially with hapi now embracing async/await.

I do not know the ideal API for this, but here's something to get started with.

```js
it('works great.', async function () {

    this.onCleanup(async () => rimraf('some/file.txt'));

    await otherTestActivities();
});
```

Another option is to loosen the restriction, and only complain if `done()` is actually called in a test that returns a promise.  Then it could be ignored as the first argument, and `onCleanup` could still be utilized.

geek:
Closes #752 
Closes #765 

* Remove support for `(done)` callback on tests
* Move to `flags` object passed into tests with `onCleanup` and `note` functions 
* Removed support for parallel mode
* Removed `debug` mode, this is now always on

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,22 +12,21 @@
   },
   "dependencies": {
     "bossy": "3.x.x",
-    "code": "4.1.x",
-    "diff": "3.3.x",
-    "eslint": "4.7.x",
+    "code": "5.1.x",
+    "diff": "3.4.x",
+    "eslint": "4.8.x",
     "eslint-config-hapi": "10.x.x",
     "eslint-plugin-hapi": "4.x.x",
-    "espree": "3.5.x",
+    "espree": "^3.5.1",
     "find-rc": "3.0.x",
     "handlebars": "4.x.x",
-    "hoek": "4.x.x",
-    "items": "2.x.x",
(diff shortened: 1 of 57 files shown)


package.json
```diff
@@ -12,22 +12,21 @@
   },
   "dependencies": {
     "bossy": "3.x.x",
-    "code": "4.1.x",
-    "diff": "3.3.x",
-    "eslint": "4.7.x",
+    "code": "5.1.x",
```
hueniverse: No no no. Time to remove it and put it back in dev deps. People should get their assertion module explicitly.

PR #786, opened 2017-11-08
Add mustCall feature

geek:
Requested by @cjihrig

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,25 @@ lab.test('attaches notes', (flags) => {
 
 Multiple notes can be appended for the same test case by simply calling `note()` repeatedly.
 
+#### `mustCall()`
+
+Declare that a particular function must be called a certain number of times. The signature to `mustCall` is `(fn, numberOfExecutions)` and it returns a wrapped copy of the `fn`. After the test is complete, each `mustCall` assertion will be checked and the test will fail if any function was called the incorrect number of times.
+
+Below is an example demonstrating how to use `mustCall` to verify that `fn` is called exactly two times.
+
+<!-- eslint-disable no-undef -->
+```javascript
+lab.test('fn must be called twice', async (flags) => {
+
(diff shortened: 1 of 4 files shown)


Marsup:
Shouldn't it be part of `code` ?

cjihrig:
This feature has two parts - first you declare what functions have to be called, then at the end of the test you have check that it was called. The problem with adding it to `code` is that you'd have to remember to explicitly check at the end of the test. This could go wrong if you forget or have a logic error in your test.

I don't know if it matters, but FWIW, this is how it's handled in Node core as well. Test writers mark functions as `mustCall`, and then it is automatically checked at the end of the test.

Marsup:
Sinon does it with 2 (or more) parameters to allow it to keep the function binding, shouldn't we be doing that ?

cjihrig:
I don't think that's necessary. [This](https://github.com/nodejs/node/blob/master/test/common/README.md#mustcallfn-exact) is the API being implemented. The function is wrapped like [this](https://github.com/continuationlabs/will-call/blob/4dab8834b0682f6645f886fc0fcdea4a4646f469/lib/index.js#L28-L31), so context and arguments are all preserved. [Here is a simple example](https://github.com/nodejs/node/blob/f002c3d2edad51bfaaf825bc6c5a8ea78fe63d16/test/parallel/test-child-process-exec-cwd.js#L37) of how this is used in Node core.

corbinu:
Anything I can do to help with this?

geek:
@corbinu thanks, I'll include this with the next release

corbinu:
@geek Great! noticed good has a test that could use this during my update

corbinu:
@geek Just checking in on if you know when the new release might be? Was going to do some more updates to good and wanted to update the tests to use this also. Totally understand we all get super busy though.

PR #797, opened 2017-12-09
Update TypeScript typings

corbinu:
I merged in my typings and updated for the new use of async/await.

I also removed the Expect typings. They can be added back in however given that other assertion libraries can be used I think there are two better options:

1) Currently I simply import expect from code separately and then use the `@typings/code` package. I have some enhancements I can simply commit to there.
2) We could merge typings into code as well and import those.

Would love to get peoples thoughts on which is better


geek:
@corbinu I'd vote for 2 to get the code typings in.

Issue #798, opened 2017-12-10
expect is not a function


jcowgar:
I've installed lab and cannot run the examples because expect is missing. Did I miss a step somewhere?

```
$ npm install --save-dev lab
$ cat > test/example.test.js
const { expect, it } = exports.lab = require('lab').script();

it('returns true when 1 + 1 equals 2', () => {

    expect(1 + 1).to.equal(2);
});
$ lab
Failed tests:

  1) returns true when 1 + 1 equals 2:

      expect is not a function

      at it (D:\Users\jerem\Projects\AMPS\dragon-js\test\example.test.js:5:5)
      at Immediate.setImmediate [as _onImmediate] (D:\Users\jerem\Projects\AMPS\dragon-js\node_modules\lab\lib\runner.js:566:31)
      at runCallback (timers.js:785:20)
      at tryOnImmediate (timers.js:747:5)
      at processImmediate [as _immediateCallback] (timers.js:718:5)


1 of 1 tests failed
Test duration: 3 ms
No global variable leaks detected
```

cjihrig:
`expect()` comes from the `code` assertion library, not `lab` itself. There was a brief time recently when they were bundled together, but now you need to install and `require('code')` to get it.

jcowgar:
Yes, that worked. So really this should be closed and another ticket created to fix the docs?

jcowgar:
Yes, that worked. I'll create a ticket, maybe a pull request, to fix the getting started docs.

PR #801, opened 2018-01-05
update to use expect from code

xdl:
Updating the examples on the readme to reflect the assertion functions being moved into code (https://github.com/hapijs/lab/issues/798)


geek:
@xdl thanks for the updates

Issue #762, opened 2017-09-27
Request to consolidate parsers


cjihrig:
There is a parser configuration for the [linter](https://github.com/hapijs/lab/blob/97cea6085264bc37eac5deceaa8707da5e4ec426/lib/linter/.eslintrc.js#L5-L7), and another parser configuration for [coverage](https://github.com/hapijs/lab/blob/3d8b49e3b6c52aef817f01575b8ce3d831d9eb7c/lib/coverage.js#L254-L257).

Is it possible to consolidate the parsers, or at least the options, in one place so that they don't become out of sync?

Marsup:
I second that one as I had this issue on object spreading being supported by coverage and not linting.

PR #802, opened 2018-01-08
Support object rest/spread under eslint

devinivy:
Object rest/spread has been supported in node since at least v8.6.0 ([here](http://node.green/#ESNEXT-candidate--stage-3--object-rest-spread-properties)), so I figured it may be appropriate to support that under our eslint config.  That is all :)

Here are the relevant eslint docs: https://eslint.org/docs/1.0.0/user-guide/configuring#specifying-language-options


cjihrig:
Somewhat related: https://github.com/hapijs/lab/issues/762

PR #808, opened 2018-02-09
add URL and URLSearchParams to global variables

mcollina:
Those will be globals in Node 10 (and current are in node master).
I'm sending this PR early because shot is currently failing on CITGM.

Ref https://github.com/hapijs/shot/pull/91.

### DIFF
--- a/lib/leaks.js
+++ b/lib/leaks.js
@@ -114,6 +114,13 @@ exports.detect = function (customGlobals) {
         whitelist.WebAssembly = true;
     }
 
+    // $lab:coverage:off$
+    if (global.URL) {
+        whitelist.URL = true;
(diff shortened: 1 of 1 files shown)


Marsup:
Should probably be added like in L92-115.

mcollina:
Done.

mcollina:
I do not know why that test is failing (see https://travis-ci.org/hapijs/lab/jobs/339389589), I think that test is flaky.

Marsup:
Can't get a flake locally but apparently travis can indeed.

geek:
Thanks for the help @mcollina

mcollina:
No worries, I'm trying to get CITGM greener, and this was on the list.

PR #811, opened 2018-02-12
Can exclude files from coverage

geek:


lib/coverage.js
```diff
@@ -49,10 +49,29 @@ exports.instrument = function (options) {
...
+    const base = internals.escape(coveragePath);
+    let excludes = options.coverageExclude ? [].concat(options.coverageExclude).map((path) => {
+
+        let isFile = false;
+        try {
+            const pathStat = Fs.statSync(Path.join(coveragePath, path));
+            isFile = pathStat.isFile();
+        }
+        // ok to swallow the error as these can be sub folders, or not even exist (which is ok)
+        catch (ex) {}
```
cjihrig: Maybe check if `ex` is an expected/acceptable error before ignore it?

lib/coverage.js
```diff
@@ -49,10 +49,29 @@ exports.instrument = function (options) {
...
+
+        return {
+            isFile,
+            escaped: internals.escape(path)
+        };
+    }) : '';
```
cjihrig: It might just be me, but the large `map()` function combined with the ternary makes this a bit harder to read.

lib/coverage.js
```diff
@@ -49,10 +49,29 @@ exports.instrument = function (options) {
...
+    excludes = excludes ? excludes.map((exclude) => {
```
cjihrig: It looks like `excludes` will either be an empty string or array at this point. If so, couldn't you move this logic up into the previous `map()`?

geek:
@cjihrig fixed up

danielo515:
How is this supposed to work ? Does it work with globs ?

danielo515:
Thabks for the clarification @geek.
Have you considered adding support for wildcard or at least regular expressions?

geek:
@danielo515 At the moment I've been fine with simply listing the folders/files to exclude directly. Is there a case when this won't work?

danielo515:
Thanks for keeping the conversation open @geek ,

As I explained on my issue #853 I have my test files along with the files being tested. This is quite convenient for visibility and easy of file requiring. The problem is that test files themseves are also being checked for coverage.
This should not be a problem normally because test are all executed, but in the case I exposed for some weird reason the coverage is not properly detected.

PR #813, opened 2018-02-22
Fail test if a falsy value is thrown by the test

### LINKED ISSUE #812: awaiting a rejected promise can still pass test
The following test "passes" in lab@15.2.2 when I would _not_ expect it to as it `awaits` a rejected promise
```JavaScript
it.only('lab should mark as fail, but doesn`t', async () => {

    let rejectP;
    const p = new Promise((resolve, reject) => {

        rejectP = reject;
    });
    setTimeout(rejectP, 5); // if I passed a param to rejectP the test would fail

    await p;

    console.log("never called - but test passes");
    throw new Error('This test should fail, but passes');
});
```
What happens is that `await p` results in an uncaught exception, but because `rejectP` was called with no parameter then all the checks like `if (err ` in the runner pass.

A quick hack for me would be to add the below `|| 'rejected'` to [lib/runner.js#L581](https://github.com/hapijs/lab/blob/master/lib/runner.js#L581)
```JavaScript
return finish(ex || 'rejected');
```

I am happy to prepare a PR if you are agreeable. Let me know.

rankida:
This fixes #812 _awaiting a rejected promise can still pass test_. All the info can be found here.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -578,7 +578,7 @@ internals.protect = function (item, state) {
                 finish();
             }
             catch (ex) {
-                return finish(ex);
+                return finish(ex || 'reject');
             }
         });
     });
--- a/test/runner.js
+++ b/test/runner.js
@@ -279,6 +279,45 @@ describe('Runner', () => {
         expect(notebook.tests[1].err.toString()).to.contain('A different reason why this test failed');
     });
 
+    it('should fail tests that throw a falsy value', async () => {
+
+        for (const falsy of ['', null, undefined]) {
+
+            const script = Lab.script({ schedule: false });
+
+            script.test('a', () => {
+
+                throw falsy;
+            });
+
+            const notebook = await Lab.execute(script, {}, null);
+            expect(notebook.tests).to.have.length(1);
+            expect(notebook.failures, 'When throwing ' + falsy).to.equal(1);
+            expect(notebook.tests[0].err.toString()).to.contain('Non Error object received or caught');
+        }
+    });
+
+    it('should fail test when returning an empty rejected promise', async () => {
+
+        const script = Lab.script({ schedule: false });
+
+        script.test('a', () => {
+
+            let rejectPromise;
+            const promise = new Promise((_, reject) => {
+
+                rejectPromise = reject;
+            });
+            setTimeout(rejectPromise, 10);
+
+            return promise;
+        });
+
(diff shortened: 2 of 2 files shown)


lib/runner.js
```diff
@@ -578,7 +578,7 @@ internals.protect = function (item, state) {
                 finish();
             }
             catch (ex) {
-                return finish(ex);
+                return finish(ex || 'reject');
```
Marsup: Shouldn't it be a `new Error()` ?

rankida: If it is an `Error` then it will flow the full way through and report an empty string as the error, whereas setting it as a string results in the standard `Non Error object received or caught` message.

I am happy enough with either (my priority is that the test fail) so really depends on what feedback makes most sense

**with 'reject'**
```
Failed tests:

  27) Executor jobs throw:

      Non Error object received or caught

      at finish...
```

**with `new Error()`**
```
   27) Executor jobs throw:



      at Immediate.setImmediate...
```

geek: This seems reasonable to me, @Marsup any objections to the current implementation?

ORESoftware:
maybe dis better?

```js
return finish(ex || 'dummy hapijs reject');
```

lib/runner.js
Marsup: OK, didn't investigate enough to tell, I was just surprised that it could end with an error or a string. If you're happy I'm happy.

lerouxb: What's wrong with `new Error('Empty promise rejection.')`?

rankida: As I said above, the "problem" with using an genuine `Error` is that you follow the same path as any errors that come from the test.

So you could see the error message `'Empty promise rejection.'` and think your test was actually throwing that.

The reason I chose a non-`Error` value is that I think the existing message `Non Error object received or caught` is the most appropriate and fits in nicely with the existing behaviour.

That said, as long as it is set to _something_ I do not care too much. My priority is that it fails the test.

rankida: @geek if you want me to change anything just let me know

rankida:
@ORESoftware thanks for your comment.

The string is used only internally to `lab` so I don't think it matters too much. If it does need changed I would prefer something that does not mention `hapi` as you might think this si something coming from the webserver.

I have addressed the comments about using `Error` in the thread above

PR #816, opened 2018-03-07
Bump diff to 3.5

Marsup:
Got an alert for a security issue on diff (https://snyk.io/vuln/npm:diff:20180305).

Unlikely lab would be the sole culprit but still, satisfying CI for worried people 🙃


AdriVanHoudt:
Node 9 fails for some reason

Marsup:
I know, it failed before I.

PR #819, opened 2018-03-12
Implement minimum assertions threshold. Closes #690.

### LINKED ISSUE #690: Allow to specify minimum number of executed assertions
Hello,

Just like you can specify the minimum code coverage percentage, I think it will be very useful to specify the minimum number of assertions to be executed per test. It have happened to me that one colleague wrote several tests, with 100% code coverage, but there were a couple of tests where 0 assertions were executed. For me this is a bug on the test, and it should be flagged as failed.

When does this happen? In a situation like this

```js
// Test shortcuts
const describe = lab.experiment;
const it = lab.test;
const expect = Code.expect;

describe('********** Example suite **********', () => {

    describe('----- Example test -----', () => {

        it('Should create an instance of a User', (done) => {

            thisReturnsAPromise() 
                        .then(()=>done()) // instead of the expected rejection, we got a valid response
                       .catch( err => { expect(err).to.be.an.insanceOf(Error); done()} )
        });
    });
});

```

The done is called to avoid timeouts.
The test pass, and they cover certain lines, but it happens that no assertion was executed. This is obviously a mistake, but that is the point of the tests, to catch up errors. If I can set a minimum of assertions to be executed to 2, that mistake would be evident.

If this is already available, there is no documentation about it.

Marsup:
For some reason I couldn't enable it for lab itself as a few tests failed, probably something to do with lab testing lab...

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Lead Maintainer: [Wyatt Preul](https://github.com/geek)
 - `--lint-fix` - apply any fixes from the linter, requires `-L` or `--lint` to be enabled. Disabled by default.
 - `--lint-options` - specify options to pass to linting program. It must be a string that is JSON.parse(able).
 - `-o`, `--output` - file to write the report to, otherwise sent to stdout.
+- `-p`, `--default-plan-threshold` - sets the minimum number of assertions a test must run. Overridable with [`plan`](#plan).
 - `-P`, `--pattern` - only load files with the given pattern in the name.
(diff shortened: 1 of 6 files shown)


geek:
Thanks, @Marsup... this will be nice to have to enable at least 1 assertion for every test. Maybe when we do v16 we can default the min-threshold to 1.

Marsup:
I don't know, it's good as an option as well, not everyone is using code.

PR #820, opened 2018-03-20
Ignore default-plan-threshold for anything else than tests

Marsup:
Follow-up to #819.

I don't think the default assertion threshold should apply to befores and afters, I didn't notice it in the initial PR but I don't think this part makes sense after using it. Would you agree ?


geek:
@Marsup absolutely agree. I'm going to keep this as 15.4.1 since the previous enhancement release is still new.

PR #824, opened 2018-03-30
Calculate relative filename using node's path

### LINKED ISSUE #823: Transforming files does not work with yarn workspaces
Hey 👋 
Got striked with rather rare error while using lab transform option - calculated relative path for the file to be transformed was looking like: `dules/babel-plugin-transform-object-rest-spread/lib/index.js`

Issue was the fact that I was using yarn workspaces and some of the packages was required from the root dir's node_modules of the repository and was used in a sub directory/workspace.

Issue was caused at https://github.com/hapijs/lab/blob/master/lib/transform.js#L63. Replacing it with [path.relative](https://nodejs.org/api/path.html#path_path_relative_from_to) solved the problem for me.

I'd be happy to provide PR fixing this issue.

pdziok:
Resolves #823

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -3,6 +3,7 @@
 // Load modules
 
 const Fs = require('fs');
+const Path = require('path');
 
 // Declare internals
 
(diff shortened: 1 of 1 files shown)


geek:
@pdziok thanks for fixing this, sorry it took so long to get released. But this is included in 15.4.3.

PR #833, opened 2018-05-13
Add WHATWG Encoding API globals

MarkTiedemann:
`TextEncoder` and `TextDecoder` may become global variables in Node 11.

I don't expect this PR to be merged now, I'm just sending this as a tiny, early heads-up. 🙂 

Issue: https://github.com/nodejs/node/issues/20365
PR: https://github.com/nodejs/node/pull/20662


geek:
@MarkTiedemann and @cjihrig this is published as 15.4.5

PR #836, opened 2018-05-17
Update coverage ECMA version to 9 (2018).

mathewdgardner:
Close if wrong approach but with Node 10 we now have async iterators / generators. While I can have eslint use ecma 2018, running lab with coverage without this change pukes with `SyntaxError: Unexpected token await` errors.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -270,7 +270,7 @@ internals.instrument = function (filename) {
         ecmaFeatures: {
             experimentalObjectRestSpread: true
         },
-        ecmaVersion: 8
+        ecmaVersion: 9
     });
 
     // Process comments


lib/coverage.js
```diff
@@ -270,7 +270,7 @@ internals.instrument = function (filename) {
         ecmaFeatures: {
```
devinivy: It's worth verifying, but I think this `ecmaFeatures` config can go away if we bump the `ecmaVersion` to `9`.

mathewdgardner: It looks like the option is removed in espree v4.0.0.alpha.0 https://github.com/eslint/espree/pull/374

geek:
I'll verify if this will break anything and then merge. Thanks for the update.

PR #838, opened 2018-06-01
Support test context from befores

geek:


Marsup:
It's a bit weird that the nested tests don't share the upper context, don't you think ?

README.md
```diff
@@ -277,6 +277,23 @@ lab.test('cleanups after test', (flags) => {
 });
 ```
 
+#### `context`
+
+`context` is an object that is passed to before and after functions in addition to tests themselves. The intent is to be able to set properties on `context` inside of a before function that can be used by a test function later. This should help reduce module level variables that are set by `before`/`beforeEach` functions. Tests aren't able to manipulate the context object for other tests.
```
cjihrig: The before and afters in this paragraph should probably be in backticks. You should probably also mention that contexts are inherited.

lib/runner.js
```diff
@@ -277,7 +279,7 @@ internals.executeExperiments = async function (experiments, state, skip) {
 
         // Sub-experiments
 
-        await internals.executeExperiments(experiment.experiments, state, skipExperiment);
+        await internals.executeExperiments(experiment.experiments, state, skipExperiment, state.currentContext);
```
cjihrig: Since this is attached to `state`, and `state` is already an argument, do you need to add a new argument?

test/index.js
```diff
@@ -222,6 +222,92 @@ describe('Lab', () => {
...
+                expect(context.test).to.not.exist();
+            });
+
+            script.afterEach(({ context }) => {
+
+                expect(context.first).to.equal(1);
+                expect(context.second).to.equal(2);
+                expect(context.third).to.equal(3);
+                expect(context.fourth).to.equal(4);
+                expect(context.test).to.not.exist();
```
cjihrig: I'm not sure about this. If the assumption is that each test has its own context, I'd expect `context.test` to exist for one of the invocations here. I guess it could be hard to track individual tests across before, beforeEach, the test, after, and afterEach. Maybe this should be documented.

Marsup:
I just hope people won't put heavy objects on that context, otherwise the cloning will be very expansive.

lib/runner.js
geek: state gets reused, so I want to avoid any confusion

test/index.js
geek: I don't want a test to be able to manipulate the context, so keeping each separate for now... we can definitely change this if we want

geek:
@Marsup yep, agreed on the clone. I'm planning to use it for very simple objects

PR #840, opened 2018-06-19
update eslint-config-hapi dependency

cjihrig:
This is a major version bump of the linting config. A lot of code is likely to need updating, particularly around the use of `padding-line-between-statements`. See https://github.com/cjihrig/eslint-config-hapi/pull/52 for all of the fun details.


hueniverse:
Can I have this?

Issue #782, opened 2017-11-01
Disable "uncaughtException" handler for tests


kanongil:
In my [exiting](https://github.com/kanongil/exiting) module I have code that triggers on `uncaughtException`, which I want to test. This works in v14, but fails in v15.

Maybe add an option to disable the built-in `uncaughtException` handling?

geek:
@kanongil can you remove all listeners before adding your listener... or remove all listeners but yours in the tests?

kanongil:
Yeah, I guess I can do a `process.removeAllListeners('uncaughtException');` before my listener is attached.

dominykas:
It would be handy to have this built-in... So that you could do something like this:

```javascript
it('rethrows a system error to crash the process', async (flags) => {

  const team = new Teamwork();
  flags.onUnhandledRejection((err) => {

    expect(err).to.be.an.error('TypeError: obj.noSuchMethod is not a function');
    team.attend();
  });

  myModule.asyncOperation();

  await team.work;
});
```

I'm thinking of releasing a helper (as suggested - by removing lab's listeners during the test) to deal with the rejections/exceptions, something like this:

```javascript
afterEach(() => processErrorHandlers.restore());

it('rethrows a system error to crash the process', async (flags) => {

    processErrorHandlers.intercept(); // records existing handlers and attaches our own

   myModule.asyncOperation();

   expect(processErrorHandlers.rejections[0]).to.be.an.error('TypeError: obj.noSuchMethod is not a function');
});
```

However, if something goes wrong, and they handlers are not reattached - then the process will crash and lab will stop intercepting the errors and things will not look pretty.

dominykas:
Now that I think of it, maybe the second approach is a nicer API, so lab could have something like

```
flags.recordUnhandledRejections(3); // after three rejections - fail the same way as it happens currently
```

and would restore/assert the behavior automatically?

geek:
@dominykas can you wrap your operating in a try/catch or create a `catch` function to trap the error?

dominykas:
@geek I suppose it's always possible to rearchitect the code in such a way, that all promises/errors thrown are handled or bubble up to a single "main" app `run()` promise, but it does get tricky under certain situations, e.g. `setInterval(async () => { /* process stuff */ })` or even a `setTimeout(() => { thrown new Error('synchronously') })`. In this situation you basically need an explicit way to "reject" the main app or if you're a library - you need to create some context and emit an error there.

[...]

PR #843, opened 2018-07-24
Allow custom handling for global errors/rejections

dominykas:
Closes https://github.com/hapijs/lab/issues/782

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -574,12 +574,27 @@ internals.protect = function (item, state) {
         /* $lab:coverage:off$ */
         const promiseRejectionHandler = function (err) {
 
+            if (flags.onUnhandledRejection) {
+                flags.onUnhandledRejection(err);
+                return;
+            }
+
             failedWithUnhandledRejection = true;
             finish(err, 'unhandledRejection');
         };
         process.on('unhandledRejection', promiseRejectionHandler);
 
         const processUncaughtExceptionHandler = function (err) {
+
+            if (flags.onUncaughtException) {
+                try {
+                    flags.onUncaughtException(err);
+                    return;
+                }
+                catch (errInErrorhandling) {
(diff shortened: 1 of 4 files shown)


test/cli_error/failure.js
```diff
@@ -47,4 +47,70 @@ describe('Test CLI', () => {
             });
         });
     });
+
+    it('passes rejection to flags.onUnhandledRejection handler', (flags) => {
```
dominykas: I figured it's best to keep all these tests together so that `// covered by test/cli_error/failure.js` comment for disabled coverage remains true?

lib/runner.js
```diff
@@ -574,13 +574,28 @@ internals.protect = function (item, state) {
...
 
         const processUncaughtExceptionHandler = function (err) {
 
+            if (flags.onUncaughtException) {
+                try {
+                    flags.onUncaughtException(err);
+                    return;
+                }
+                catch (errInErrorhandling) {
+                    err = errInErrorhandling;
```
dominykas: I figure if the user has `onUncaughtException` declared, then they know what they expect as an `err`, so it's best to fail the test with whatever happens inside of it, rather than with the original error.

lib/runner.js
```diff
@@ -574,13 +574,28 @@ internals.protect = function (item, state) {
         /* $lab:coverage:off$ */
         const promiseRejectionHandler = function (err) {
 
+            if (flags.onUnhandledRejection) {
+                flags.onUnhandledRejection(err);
```
dominykas: If this throws - it's handled by `onUncaughtException` - I think `try/catch` is not necessary here?

lib/runner.js
geek: This sounds like a good plan to me

PR #846, opened 2018-07-29
Only is no longer limited to 1 test

geek:


test/index.js
```diff
@@ -513,6 +513,47 @@ describe('Lab', () => {
...
+        });
+
+        script.experiment('test2', {}, () => {
+
+            script.test.only('works', () => {});
+        });
+
+        const notebook = await Lab.execute(script, null, null);
+        expect(notebook.tests).to.have.length(2);
+        //console.log(notebook.tests)
```
cjihrig: 🙈

Marsup:
Did it work after all?

geek:
yes, you can have multiple only's at different levels and everything else is skipped.

Marsup:
When we discussed it a few months back, I mentioned a bug, it's still present unless I misunderstood the feature. You can see it with https://github.com/hapijs/joi/commit/da9355e14bf8ef60aae71132c3fc3778a1b39b7d, it runs no test at all.

geek:
@Marsup here is a fix: https://github.com/hapijs/lab/pull/867

Marsup:
Seems to work, superb! I'm not sure about those test failures but it fixes my case.

PR #847, opened 2018-07-31
Changed the classname and the name in the junit report generated to t…

ShouriGupta:
…est suite name and test case name respectively.

### DIFF
--- a/lib/reporters/junit/report.xml
+++ b/lib/reporters/junit/report.xml
@@ -1,6 +1,6 @@
 <testsuite tests="{{this.count}}" skipped="{{this.skipped}}" errors="{{this.errors}}" failures="{{this.failures}}" time="{{this.time}}">
     {{#each this.tests~}}
-    <testcase classname="{{this.title}}" name="{{this.title}}" time="{{this.time}}">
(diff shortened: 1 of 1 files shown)


lib/reporters/junit/report.xml
```diff
@@ -1,6 +1,6 @@
 <testsuite tests="{{this.count}}" skipped="{{this.skipped}}" errors="{{this.errors}}" failures="{{this.failures}}" time="{{this.time}}">
     {{#each this.tests~}}
-    <testcase classname="{{this.title}}" name="{{this.title}}" time="{{this.time}}">
+    <testcase classname="{{this.path}}" name="{{this.relativeTitle}}" time="{{this.time}}">
```
geek: Why should the name change to the `relativeTitle`? If we add more wrapping testsuite elements for each of the experiments then maybe we can reduce the name for each testcase.

ShouriGupta: relativeTitle is the test case name. As per junit standards we can have classname as suite name and name as testcase name.

geek:
Marking as a breaking change in case someone is relying on the existing test names in the lab junit output

PR #871, opened 2018-10-30
Improve console coverage report readability

mkantor:
The way coverage reports were presented in the console made them hard to skim. I made a few tweaks to try and improve this:

- Report runs of contiguous lines as ranges (e.g. "7-11") instead of listing every line number.
- Truncate coverage reports when they get too long.
- Color the detail portion yellow instead of red.

### Before:
screen shot 2018-10-30 at 3 36 28 pm

### After:
screen shot 2018-10-30 at 3 38 46 pm

Let me know what you think!


mkantor:
Thanks for merging!

Issue #710, opened 2017-05-29
setting coverage-path doesn't include all code in path


danielb2:
I'm setting coverage-path to `--coverage-path lib`. As a result, I'm expecting ALL files in the `lib` dir to be part of the coverage report.

Actual behavior:
Files not required in the test files are not covered. 

https://github.com/danielb2/lab-example

It would also be useful to be able to specify multiple paths to include coverage for

geek:
I think this a reasonable idea, any files under the coverage-path will be traversed and required. It will be breaking as coverage percentages might change and break existing builds. I wonder if we should add an option for opting-in or out of this auto-require behavior.

jcollum-cambia:
My 2 cents: automatic with opt-out. We were just discussing why we have to manually exclude code in ./test path but the code in my ./src path isn't automatically included in coverage.

joshrivers:
What if we added a `--coverage-all-path` flag, leaving `--coverage-path` with it's current behavior. Later iterations could deprecate --coverage-path and migrate users forward?

PR #877, opened 2018-11-02
Add --coverage-all option

ajgamble-milner:
Fixes https://github.com/hapijs/lab/issues/710

- Adds option to specify `--coverage-all` flag.
- Reports coverage for all files that match `--coverage-path`, regardless of a test touching them
- Respects `--coverage-exclude`
- Adds `--coverage-flat` to prevent recursive finding of files for coverage

One outstanding issue, which I am currently unsure how to resolve:
```
Coverage: 99.96% (1/2254)
lib/coverage.js missing coverage on line(s): 412
Code coverage below threshold: 99.96 < 100
```

which refers to: `if (stat.isFile() && Path.basename(nextPath)[0] !== '.') {` 🤷‍♀️ 

Feedback on the changes welcome and a suggestion of how to hit the code coverage would be helpful :+1:

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -207,6 +207,16 @@ internals.options = function () {
             multiple: true,
             default: null
         },
+        'coverage-all': {
+            type: 'boolean',
+            description: 'include all files in coveragePath in report',
+            default: null
+        },
+        'coverage-flat': {
+            type: 'boolean',
+            description: 'prevent recursive inclusion of all files in coveragePath in report',
+            default: null
+        },
         'default-plan-threshold': {
             alias: 'p',
             type: 'number',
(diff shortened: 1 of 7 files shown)


lib/cli.js
```diff
@@ -423,7 +433,7 @@ internals.options = function () {
     options.paths = argv._ ? [].concat(argv._) : options.paths;
 
     const keys = ['assert', 'bail', 'colors', 'context-timeout', 'coverage', 'coverage-exclude',
-        'coverage-path', 'default-plan-threshold', 'dry', 'environment', 'flat', 'globals', 'grep',
+        'coverage-path', 'coverage-all', 'coverage-flat', 'default-plan-threshold', 'dry', 'environment', 'flat', 'globals', 'grep',
```
geek: Somewhere later in the `internals.options` function can you also add a check for if `coverage-flat` and not `coverage-all` then log an error and exit. Since the `coverage-flat` requires the `coverage-all` flag I want to make sure we catch the case when its not provided. Or do you think we should set `--coverage-all` if `--coverage-flat` is provided?

lib/coverage.js
```diff
@@ -385,6 +385,50 @@ internals.instrument = function (filename) {
...
+            const stat = Fs.statSync(nextPath);
+            if (stat.isDirectory() &&
+                !options['coverage-flat']) {
+
+                files = files.concat(traverse(nextPath, options));
+                return;
+            }
+
+            if (stat.isFile() &&
+                Path.basename(nextPath)[0] !== '.') {
```
geek: In the `test/coverage/coverage-all/nested-folder/` you can try adding a `.dir/somefile.js` and verify that this case is tested

lib/cli.js
ajgamble-milner: I think the error and exit would be the correct way to go, it should be explicit with what's going on :+1:

lib/coverage.js
ajgamble-milner: The coverage path was incorrect to fully test this scenario, updated!

lib/cli.js
ajgamble-milner: I have also added to the options to enable coverage when `--coverage-all` is set, in a similar behaviour to `--threshold` and some of the reporters

PR #878, opened 2018-11-02
Fix false-positive coverage of if statements (REDUCES REPORTED COVERAGE!)

rluba:
Lab currently reports the last part of a conditional within an if statement as covered even though one case might not have been covered!

For example, the following `if` statement is reported as fully covered, even if `value2` is always `false`:
```JavaScript
if (value1 || value2) {
…
}
```

**This PR causes the coverage of Lab itself to drop** because it exposes several uncovered conditionals within Lab  that have previously been incorrectly reported as covered.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -242,7 +242,7 @@ internals.instrument = function (filename) {
         }
         else if (node.type === 'LogicalExpression') {
             const left = addStatement(line, node.left, true);
-            const right = addStatement(line, node.right, node.parent.type === 'LogicalExpression');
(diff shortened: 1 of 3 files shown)


geek:
@rluba I'm [working through](https://github.com/geek/lab/tree/cov) the missing tests and will PR into your PR so that the merge doesn't break master.

rluba:
> In the assignment case, you don't want to test if y is always true or false because it doesn't matter.

@hueniverse Exactly, that’s why I left it as it is for the assignment case. But ignoring the last boolean in an `if` clause caused us to miss some serious coverage issues.

PR #882, opened 2018-11-07
Add missing coverage and remove unnecessary checks

geek:
Adds missing coverage discovered from https://github.com/hapijs/lab/pull/878

### DIFF
--- a/lib/leaks.js
+++ b/lib/leaks.js
@@ -170,8 +170,7 @@ exports.detect = function (customGlobals) {
     const leaks = [];
     const globals = Object.getOwnPropertyNames(global);
     for (let i = 0; i < globals.length; ++i) {
-        if (!whitelist[globals[i]] &&
-            global[globals[i]] !== global) {
+        if (!whitelist[globals[i]]) {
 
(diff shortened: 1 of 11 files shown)


lib/reporters/console.js
```diff
@@ -354,7 +354,7 @@ internals.Reporter.prototype.end = function (notebook) {
                 }
             });
 
-            if (coverage.percent < this.settings.threshold || isNaN(coverage.percent)) {
```
geek: NaN cannot make it into this branch because of the conditional check above `coverage.percent < 100`

lib/leaks.js
```diff
@@ -170,8 +170,7 @@ exports.detect = function (customGlobals) {
     const leaks = [];
     const globals = Object.getOwnPropertyNames(global);
     for (let i = 0; i < globals.length; ++i) {
-        if (!whitelist[globals[i]] &&
-            global[globals[i]] !== global) {
+        if (!whitelist[globals[i]]) {
```
hueniverse: No need for the extra new lines...

PR #897, opened 2018-11-20
Add '--mutable-context' option to allow tests to mutate context.

Invader444:
Currently I have been executing this hack in `before(({ context }) => ... )` to allow mutable context...
```
    /*
     * @HACK
     * Allow the lab context to be mutable across tests by fooling Hoek.clone(context) into thinking we are Immutable.js...
     * Hopefully upstream makes this a config option at some point...
     */
    const labContextPrototype = Object.getPrototypeOf(context)
    if (!labContextPrototype) {
        Object.setPrototypeOf(context, {
            isImmutable: () => false
        })
    } else if (!labContextPrototype.isImmutable) {
        Object.setPrototypeOf(context, {
            ...labContextPrototype,
            isImmutable: () => false
        })
    }
```

having an option to explicitly enable tests to mutate context would be more ideal! :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Lead Maintainer: [Wyatt Preul](https://github.com/geek)
 - `--lint-options` - specify options to pass to linting program. It must be a string that is JSON.parse(able).
 - `-m`, `--timeout` - individual tests timeout in milliseconds (zero disables timeout). Defaults to 2 seconds.
 - `-M`, `--context-timeout` - default timeouts for before, after, beforeEach and afterEach in milliseconds. Disabled by default.
+- `--mutable-context` - allow tests to mutate flags.context. Disabled by default.
 - `-o`, `--output` - file to write the report to, otherwise sent to stdout.
 - `-p`, `--default-plan-threshold` - sets the minimum number of assertions a test must run. Overridable with [`plan`](#plan).
 - `-P`, `--pattern` - only load files with the given pattern in the name.
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -313,6 +313,11 @@ internals.options = function () {
             description: 'linter warnings threshold in absolute value',
             default: null
         },
+        'mutable-context': {
+            type: 'boolean',
+            description: 'allow tests to mutate flags.context',
+            default: null
+        },
         output: {
             alias: 'o',
             type: 'string',
@@ -405,6 +410,7 @@ internals.options = function () {
         'lint-fix': false,
         'lint-errors-threshold': 0,
         'lint-warnings-threshold': 0,
+        'mutable-context': false,
         paths: ['test'],
         reporter: 'console',
         shuffle: false,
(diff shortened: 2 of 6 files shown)


Marsup:
This looks very much like an anti-pattern, a test should never ever depend on a preceding test nor should it set things up for the next. Can't you use the existing before/beforeEach/describe nesting for this?

geek:
@Invader444 thank you for the PR and the idea for this feature. I do agree with @Marsup, mainly because we do try and make sure each test can run in isolation and doesn't depend on any test around it. I know that there are likely edge cases where being able to change the context is useful between tests, but it's not something that I feel comfortable supporting directly in lab in this way. There are also ways to have a shared context outside of lab. Another option is what @Marsup suggested with the nested before/describe

Invader444:
Suppose you have a simple game:
```
class Game {
    onAttack(attacker, defender) {
        attacker.onAttack()
        defender.onDefend()

        // Do other stuff
    }

    canAttack(attacker) {
        return this.phase === 'ATTACK_PHASE';
    }

    drawCards() {
        // Draw some cards, and then
        this.phase = 'ATTACK_PHASE';
    }
}
```

To test that a Player.onAttack() works, we first must execute Game.drawCards();

If we setup a new game in context, the game state must either be mutable, or we must have an experiment for each test, with a before that advances the game to the proper place.

Seems like a ton of useless boilerplate that way :(

Invader444:
The alternative then would be to do things the way we did before upgrading lab to a version with context, which was to have a `const game = new Game(...)` outside of each `lab.experiment(...)` :(

geek:
@Invader444 I will usually use something like https://www.npmjs.com/package/stand-in to replace one of the protos if I need it to return something different or to set `this.phase` in the example. Either that or you can have extra code to get the game to correct state and then test the `Player.onAttack()`. We tend to go with the second approach in hapi and have more code in each test instead of requiring any order to the test execution.

Invader444:
@geek, how about this example?

```
const events = {}
class Game {
    performAttack(attacker, defender) {
        this.emit('attack', {
            attacker,
            defender
        })
    }

    on(eventName, listener, callback) {
        events[eventName] = {
            listener,
            callback
        }
    }

    emit(eventName, data) {
        events[eventName].callback.apply(events[eventName].listener, data)
    }
}

class Player() {
    constructor() {
        this.points = 0
    }

    getPoints() {
        return this.points
    }
}
```

```
lab.experiment('Internal event system works', () => {
    lab.beforeEach((flags) => {
        const player1 = new Player()
        const player2 = new Player()
        const game = new Game()

        game.on('attack', player1, () => { this.points++ })

        flags.context.game = game
        flags.context.player1 = player1
        flags.context.player2 = player2

        expect(player1.getPoints()).to.equal(0)
    })

    lab.test('Attacking increments player one\'s points', ({ context: { game, player1, player2 } }) => {
        game.performAttack(player1, player2)
        expect(player1.getPoints()).to.equal(1) // FAILS
    })
})
```

geek:
@Invader444 I believe you will want to change the `() => ` so that `this` applies to the player (https://devdocs.io/javascript/functions/arrow_functions)

```
game.on('attack', player1, function () { this.points++ })
```

Issue #572, opened 2016-04-30
No mismatch highlighted on -0 !== 0 deep equality mismatch


rreusser:
**Version:** 10.3.2

**Expected behavior**: The reporter highlights all deep equality mismatches that caused the assertion to fail, allowing the user to determine the cause.

**Observed behavior**: [Hoek returns false when comparing 0 to -0](https://github.com/hapijs/hoek/blob/master/lib/index.js#L271-L273), but the reporter [uses string comparison](https://github.com/hapijs/lab/blob/master/lib/reporters/console.js#L176-L177) that converts `-0` to `"0"`, resulting in a failed assertion with no highlighted output. (side note: I didn't expect Hoek to be this strict since, generally speaking in JS, `0 === -0`)

**Steps to reproduce**: (See full gist here: [zero-test.js](https://gist.github.com/rreusser/afde9a775f5ffec8b2faa91e040e6ac9))

``` javascript
describe('negative zero check', function () {
  it('highlights mismatch when deep equality fails', function (done) {
    expect([0]).to.deep.equal([-0]);
    done();
  }); 
});
```

I get the output:

```
Failed tests:
  1) negative zero check highlights mismatch when deep equality fails:
      actual expected
      [
        0
      ]
      Expected [ 0 ] to equal specified value
      at /Users/rreusser/projects/node/zerotest/index.js:10:25
```

Of course there's no coloring here, but there's no indication of _why_ the test failed—which turns out to take a bit of time and sanity-check exhaustion to determine in a nontrivial case.

It seems like reliance on [StringifySafe](https://github.com/isaacs/json-stringify-safe) is the source of the problem since `-0` is stringified as `"0"` instead of `"-0"` so that a string diff returns nothing. Unfortunately, I can't think of a great solution here since StringifySafe otherwise does exactly what it needs to do and saves a lot of recursion and complexity—and is probably correct to convert `-0` to `"0"`. This could all be obscure enough to not be a big deal, but I'm glad to hear or offer thoughts!

Issue #914, opened 2019-04-29
"console" reporter crashes if object contains a BigInt


kanongil:
The following test crashes the `console` reporter:

```js
it('test', () => {

    expect(BigInt(1)).to.equal(BigInt(2));
});
```

Output:
```
(node:4086) UnhandledPromiseRejectionWarning: TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at stringify (/Users/gil/Development/node-m3u8parse/node_modules/json-stable-stringify/index.js:36:25)
    at module.exports (/Users/gil/Development/node-m3u8parse/node_modules/json-stable-stringify/index.js:70:7)
    at Object.internals.stringify (/Users/gil/Development/node-m3u8parse/node_modules/@hapi/lab/lib/reporters/console.js:16:12)
    at module.exports.internals.Reporter.internals.Reporter.end (/Users/gil/Development/node-m3u8parse/node_modules/@hapi/lab/lib/reporters/console.js:186:42)
    at module.exports.internals.Reporter.reporter.finalize (/Users/gil/Development/node-m3u8parse/node_modules/@hapi/lab/lib/reporters/index.js:94:28)
    at Object.exports.report (/Users/gil/Development/node-m3u8parse/node_modules/@hapi/lab/lib/runner.js:124:21)
(node:4086) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:4086) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

PR #915, opened 2019-05-01
Use util.inspect() for console reporter

kanongil:
I don't know how much sense this patch makes, but I created it to solve #914.

Basically, I have replaced the serialization of the reported object / diff. Instead of using `JSON.stringify()`, I use the node built-in `util.inspect()`.

The existing JSON-based serialization is not ideal, and is not able to represent modern types in a reasonable manner. E.g. the current implementation doesn't support `Map` and `Set` (represented as `{}`), or `BigInt` (crashes).

Using `util.inspect()` have the great advantage, that it can serialize all native types into somewhat reasonable representations, and should be updated with any new types as they become available.

One issue with this, is that node@8 doesn't support the `compact` and `sorted` options, resulting in 2 test failures.

### DIFF
--- a/lib/reporters/console.js
+++ b/lib/reporters/console.js
@@ -1,9 +1,8 @@
 'use strict';
 
+const Util = require('util');
 
 const Diff = require('diff');
-const StringifySafe = require('json-stringify-safe');
-const StableStringify = require('json-stable-stringify');
 
 
 const internals = {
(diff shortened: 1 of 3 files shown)


kanongil:
Btw. this seems to fix #572.

geek:
@kanongil seems reasonable to me. I guess the big break will be for node 8 output. Seems like a breaking change, is that correct?

hueniverse:
Not really breaking because the changes are to the console output which is not an API, but marking as such because node 8 will suffer a tiny bit with objects no longer broken into lines.

PR #923, opened 2019-05-23
Support multiple "only" flags

tpburch:
Multiple "only" flags currently work, but only in certain cases
depending on experimnet depth. This PR implements full support for
multiple "only" flags, regardless of same experiment, different
experimnt, different script, and depth.

As an approach, prior to marking non-only tests as skipped, we now mark
all parents of an "only" element as "only" (unless they are explicitly
skipped).

Also, fixed a few incorrect tests.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -150,6 +150,11 @@ exports.execute = async function (scripts, options, reporter) {
     if (onlyNodes.length) {
         onlyNodes.forEach((onlyNode) => {
 
+            internals.markParentsAsOnly(onlyNode.experiment);
+        });
+
+        onlyNodes.forEach((onlyNode) => {
+
             internals.skipAllButOnly(scripts, onlyNode);
         });
     }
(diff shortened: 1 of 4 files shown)


lib/runner.js
```diff
@@ -148,6 +148,11 @@ exports.execute = async function (scripts, options, reporter) {
 
     const onlyNodes = Hoek.flatten(scripts.map((script) => script._onlyNodes));
     if (onlyNodes.length) {
+        onlyNodes.forEach((onlyNode) => {
+
+            internals.markParentsAsOnly(onlyNode.experiment);
```
geek: Nice!

lib/runner.js
```diff
@@ -247,6 +252,19 @@ internals.enableSkip = (element) => {
     element.options.skip = true;
 };
 
+internals.markParentsAsOnly = (element) => {
+
+    if (!element.options.skip) {
```
geek: Should we assert in this case and inform the developer that they must either remove the `skip` or `only` in the experiment tree?

tpburch: I think there is a use case where the dev would do this on purpose. If multiple tests are marked as `only`, but for a given test run they want to skip some of them. Skipping a common parent experiment can be a lot quicker than deleting potentially several `only`s. This can be pretty handy when you are working with many levels of experiments.

Having said that, I can also see potential for some confusion as to why a test marked with `only` isn't running if the dev doesn't notice that it was skipped higher up. I think that case would end up being similar to the older approach of throwing when there are multiple `only`s. 

Maybe a section in the docs explaining the logic?

tpburch:
@geek @hueniverse Is there any action I need to take before this can be merged?

geek:
@tpburch I don't think there is any other changes you need to make. I need @hueniverse to merge/publish as I don't have permissions to publish.

PR #926, opened 2019-06-09
Implement coverage bypass stack

### LINKED ISSUE #904: Add affordance for coverage flag stack
If I configure a transform using `-T` with `babel`, I might want to have babel comment out portions of the code it generates with `auxiliaryCommentBefore` and `auxiliaryCommentAfter` so that `lab` doesn't try to lint the generated code. Separately, I might sometimes want to disable coverage for actual application code. If I put the two together, it's possible that a block of code that has linting disabled gets re-enabled by the generated code's comments:

```js
// input:
/* $lab:coverage:off$ */
const { types } = Util;
export const isSet = (types && types.isSet) || internals.isSet;
/* $lab:coverage:on$ */



// output:
/* $lab:coverage:off$ */
const {
  types
} =
/*$lab:coverage:off$*/
_util
/*$lab:coverage:on$*/
.
/*$lab:coverage:off$*/
default
/*$lab:coverage:on$*/
;
const isSet = types && types.isSet || internals.isSet;        // This line gets linted, even though it shouldn't have been.
/* $lab:coverage:on$ */
```

I'm curious whether there's an existing solution to this problem.

If not, would we be open to adding support for `/* $lab:coverage:push$ */` and `/* $lab:coverage:pop$ */` comments (name unimportant), which would create a running stack of flag states that auto-generated code could use to avoid mucking with the user's intended coverage state?

skeggse:
Fixes #904.

### DIFF
--- a/lib/coverage.js
+++ b/lib/coverage.js
@@ -269,19 +269,41 @@ internals.instrument = function (filename) {
 
     let skipStart = 0;
     let segmentSkip = false;
+
+    // In addition to maintaining the current coverage bypass state, support
+    // comments that push that state to a stack, allowing transpiled code to
+    // skip coverage without interfering with a user-defined bypass block.
+    const skipStack = [];
     tree.comments.forEach((comment) => {
 
-        const directive = comment.value.match(/^\s*\$lab\:coverage\:(off|on)\$\s*$/);
+        const directive = comment.value.match(/^\s*\$lab\:coverage\:(off|on|push|pop)\$\s*$/);
         if (directive) {
-            const skip = directive[1] !== 'on';
-            if (skip !== segmentSkip) {
-                segmentSkip = skip;
-                if (skip) {
-                    skipStart = comment.range[1];
+            const command = directive[1];
+            if (command === 'push') {
+                skipStack.push(segmentSkip);
+            }
+            else {
+                let newSkipState;
+                if (command === 'pop') {
+                    if (!skipStack.length) {
+                        throw new Error('unable to pop coverage bypass stack');
+                    }
+
+                    newSkipState = skipStack.pop();
                 }
                 else {
-                    for (let i = skipStart; i < comment.range[0]; ++i) {
-                        bypass[i] = true;
+                    newSkipState = command !== 'on';
+                }
+
+                if (newSkipState !== segmentSkip) {
+                    segmentSkip = newSkipState;
+                    if (newSkipState) {
+                        skipStart = comment.range[1];
+                    }
+                    else {
+                        for (let i = skipStart; i < comment.range[0]; ++i) {
+                            bypass[i] = true;
+                        }
                     }
                 }
             }
--- a/test/coverage/bypass-stack-alt.js
+++ b/test/coverage/bypass-stack-alt.js
@@ -1,0 +1,37 @@
+'use strict';
+
+// Load modules
+
+
+// Declare internals
+
+const internals = {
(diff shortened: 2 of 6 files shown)


lib/coverage.js
```diff
@@ -269,19 +269,41 @@ internals.instrument = function (filename) {
...
-                    skipStart = comment.range[1];
+            const command = directive[1];
+            if (command === 'push') {
+                skipStack.push(segmentSkip);
+            }
+            else {
+                let newSkipState;
+                if (command === 'pop') {
+                    if (!skipStack.length) {
+                        throw new Error('unable to pop coverage bypass stack');
```
skeggse: There are a number of other reasonable behaviors here, such as just resetting to the default (`on`), or doing nothing. I lean towards this behavior because it seems more helpful if you mis-implement machine generated stack manipulation comments, though in that case it may be prudent to also warn if the stack isn't empty by the end of the file. Thoughts?

lib/coverage.js
```diff
@@ -269,19 +269,41 @@ internals.instrument = function (filename) {
...
 
-        const directive = comment.value.match(/^\s*\$lab\:coverage\:(off|on)\$\s*$/);
+        const directive = comment.value.match(/^\s*\$lab\:coverage\:(off|on|push|pop)\$\s*$/);
         if (directive) {
-            const skip = directive[1] !== 'on';
-            if (skip !== segmentSkip) {
-                segmentSkip = skip;
-                if (skip) {
-                    skipStart = comment.range[1];
+            const command = directive[1];
```
skeggse: This whole block ends up being a bit of a gross deeply nested block. I could split bits out into functions or reorganize using `switch` or something; let me know if that's important - this seemed sufficient.

lib/coverage.js
```diff
@@ -269,19 +269,41 @@ internals.instrument = function (filename) {
...
+                    newSkipState = command !== 'on';
+                }
+
+                if (newSkipState !== segmentSkip) {
+                    segmentSkip = newSkipState;
+                    if (newSkipState) {
+                        skipStart = comment.range[1];
+                    }
+                    else {
+                        for (let i = skipStart; i < comment.range[0]; ++i) {
```
skeggse: it's just occurring to me that a `$lab:coverage:off$` comment without a subsequent `on` comment might result in no bypass - is this intended behavior, or just a property of how this was originally structured?

test/coverage/bypass-stack-alt.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
skeggse: I was going to have `bypass-stack.js` also have this functionality and let `getIsSet` be parameterized to the two `utilModule` copies, but I didn't immediately see how to reset the coverage state of the module.

skeggse:
Travis failures seem windows-related and unrelated to proposed changes.

geek:
@skeggse nice work! Do you think it is worth explaining how this works in the readme?

skeggse:
> @skeggse nice work! Do you think it is worth explaining how this works in the readme?

Ah yeah that seems reasonable. I'll write some docs soon.

skeggse:
Added some documentation. Does that align with your expectations?

geek:
@skeggse looks good, going to be published at @hapi/lab as 19.1.0

skeggse:
Heh I'm just noticing that I wrote the original documentation for the coverage directives in #441

PR #954, opened 2019-12-20
TS: Uses an inferred generic on flags.mustCall

ggoodman:
Previously, the type of the function returned by `flags.mustCall` did not match the type of the function being passed in. As a result, attempts to use the function returned by `flags.mustCall` would lack type safety.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -380,7 +380,7 @@ declare namespace script {
          * 
          * @returns a wrapped function.
          */
-        mustCall(func: Function, count: number): Function;
+        mustCall<T extends (...args: any[]) => any>(func: T, count: number): T;
 
         /**
          * Adds notes to the test log.


lib/index.d.ts
```diff
@@ -380,7 +380,7 @@ declare namespace script {
          * 
          * @returns a wrapped function.
          */
-        mustCall(func: Function, count: number): Function;
+        mustCall<T extends (...args: any[]) => any>(func: T, count: number): T;
```
ggoodman: This *might* impose some version constraints on the typescript version that can consume the typing info.

hueniverse: We only support latest TS versions. Is this a breaking change or not because it wasn't working to begin with?

ggoodman: Hi, sorry I didn't get back earlier. I agree with your assessment; this should be treated as a bugfix IMHO because it produces a more correct typing result.

PR #955, opened 2019-12-20
TS: String object type -> string literal type

ggoodman:
I suspect the use of the `String` object type was an accident of history. This PR makes moves to a more consistent use of types for literal types.

> Note: there is some [weirdness about type compatibility between `String` and `string`](https://www.typescriptlang.org/play/#code/C4TwDgpgBAhlC8UDKwBOBLAdgcyhAHsBJgCYDOUZaWuA-FAOQgRkNQBcjmA9gwNwAoUJCgAjBJWo48hYuWRS6jZqw5deg4dADGElBmkEipCvppR6TFm04Me-IA).

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -308,26 +308,26 @@ declare namespace script {
 
     interface Experiment {
 
-        (title: String, content: () => void): void;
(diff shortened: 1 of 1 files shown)


hueniverse:
Is this a breaking change in TS land?

Issue #977, opened 2020-03-27
Replace Handlebars with Template Strings


aorinevo:
#### Support plan

* *which support plan is this issue covered by?* Community
* *is this issue currently blocking your project?* no
* *is this issue affecting a production system?* yes

#### Context

* *node version*: 12.16.1
* *module version*: 22.0.4
* *environment* (e.g. node, browser, native): node
* *used with* hapi
* *any other relevant information*: N/A

#### What problem are you trying to solve?

Lab depends on Handlebars to generate HTML for coverage reports.  While Handlebars is a great templating language, it seems to have fallen out of maintenance.  There are currently almost 200 vulnerabilities ranging in severity from critical to low (albeit the overwhelming majority are low).  In addition, some advisories just cannot be resolved without resorting to forced resolutions or exceptions via `nsprc`.  That is, `npm audit --fix` or `npm update --depth={{depth}}` do not work.

<img width="959" alt="Screen Shot 2020-03-26 at 9 35 46 PM" src="https://user-images.githubusercontent.com/3240137/77712298-c6948600-6fa9-11ea-9064-3c0cfe012538.png">

<img width="569" alt="Screen Shot 2020-03-26 at 9 29 35 PM" src="https://user-images.githubusercontent.com/3240137/77712174-74ebfb80-6fa9-11ea-8a99-d6af78f0be11.png">

With a relatively low-level of effort, we can remove handlebars dependency altogether by using template strings.  For example, see  https://github.com/hapijs/lab/pull/978/files.

### ROI
- removes 1 dependency (~80KB, includes compiler and runtime)
- reduces install time
- resolves existing and future Handlebars vulnerabilities
- increased testability of html reporter _templates_
- unlocks potential to remove all html files (i.e. `reporters/html/partials/*.html` and `reporters/reporter.html`

### Notes
- There is an open PR on Handlebars to resolve some existing Handlebars vulnerabilities (see https://github.com/wycats/handlebars.js/pull/1662)
- This refactor was largely inspired by https://medium.com/your-majesty-co/frameworkless-javascript-template-literals-the-best-thing-since-sliced-bread-d97f000ce955
- Some consideration needed for when passing user inputs into template strings (see comment in above medium article).  Not applicable here as html reports are static.
- Maybe replace handlebars with VueJS v3.  Bundle size is estimated to be ~10KB.  
[...]

PR #978, opened 2020-03-27
feat: refactor handlebars into template literals.

aorinevo:
# Changes
- replace `handlebards` with natively supported template strings in a backwards compatible fashion.
- feature is hidden behind a feature flag.  To enable simply set `enableTemplateStringHTMLReporter` to `'true'` in `.labrc.js`.
- add lint to scripts.

## Notes
- dependent on approval of open issue https://github.com/hapijs/lab/issues/977
- while this PR adds 1069 lines of code, ~185 correspond to unit tests.  In addition, 714 lines of code can be removed once we/if backwards compatibility is dropped.  More related stats:  
  - `hapijs:master` branch consists of 95.5% javascript, 3.8% html, and 0.7% other.
  - `aorinevo:replace_handlebars` branch consists of 95.5% javascript, 3.9% html, and 0.7% other.
  - Percentage of html can go down to 0% once/if backwards compatibility is dropped.
- handlebars package can be removed from `package.json` once/if backwards compatibility is dropped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,7 @@
 - [Changelog](https://hapi.dev/family/lab/changelog/)
 - [Project policies](https://hapi.dev/policies/)
 - [Free and commercial support options](https://hapi.dev/support/)
+
+
+# Notes
+- Replacing Handlebars with template strings.  Implement similar to https://medium.com/your-majesty-co/frameworkless-javascript-template-literals-the-best-thing-since-sliced-bread-d97f000ce955
--- a/lib/reporters/html/partials/scripts.js
+++ b/lib/reporters/html/partials/scripts.js
@@ -1,0 +1,89 @@
+'use strict';
+
+exports.scripts = () => {
+
+    return `<script>
+  headings = [];
+
+  onload = function () {
+      headings = document.querySelectorAll('h2');
+      reset();
+  };
+
+  onscroll = function (e) {
+      var heading = find(window.scrollY);
+      if (!heading) {
+        var current = document.querySelector('#menu a.active');
+        if (current) {
+          current.className = '';
+        }
+        return;
+      }
+      var links = document.querySelectorAll('#menu a')
+              , link;
+
+      for (var i = 0, len = links.length; i < len; ++i) {
+          link = links[i];
+          link.className = link.getAttribute('href') == '#' + heading.id ? 'active' : '';
+      }
+  };
+
+  function find (y) {
+      var i = headings.length
+              , heading;
+
+      while (i--) {
+          heading = headings[i];
+          if (y >= heading.offsetTop) {
+              return heading;
+          }
+      }
+  }
+
+  function toggle (className) {
+
+      var elements = document.getElementsByClassName(className);
+
+      for (var i = 0, il = elements.length; i < il; ++i) {
+          var element = elements[i];
+
+          if (element.classList.contains('hide')) {
+              element.classList.remove('hide');
+              element.classList.add('show');
+          }
+          else {
+              element.classList.remove('show');
+              element.classList.add('hide');
+          }
+      };
+  }
+
+  function reset () {
+
+      var shownElements = document.getElementsByClassName('show');
+      var filterElements = document.querySelectorAll('input[type=checkbox]');
+
+      for (var i = 0, il = filterElements.length; i < il; ++i) {
+          filterElements[i].checked = false;
+      }
+
+      // Check any filters with visible elements
+      for (i = 0, il = shownElements.length; i < il; ++i) {
+          var shownElement = shownElements[i];
+          var classNames = shownElement.className.split(' ');
+          for (var ci = 0, cl = classNames.length; ci < cl; ++ci){
+              var element = document.getElementById('show-' + classNames[ci]);
+              if (element) {
+                  element.checked = true;
+              }
+          }
+      }
+  };
+
+  function filter (element) {
+
+      toggle(element.value);
+      reset();
+  }
+</script>`;
+};
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
         "eslint": "6.x.x",
         "find-rc": "4.x.x",
         "globby": "10.x.x",
-        "handlebars": "4.x.x",
+        "handlebars": "^4.7.3",
         "seedrandom": "3.x.x",
         "source-map": "0.7.x",
         "source-map-support": "0.5.x",
@@ -44,7 +44,8 @@
     },
     "scripts": {
         "test": "node ./bin/_lab -f -L -t 100 -m 10000 --types-test test/index.ts",
-        "test-cov-html": "node ./bin/_lab -f -L -r html -m 10000 -o coverage.html"
+        "test-cov-html": "node ./bin/_lab -f -L -r html -m 10000 -o coverage.html",
+        "lint": "node ./bin/_lab -d -f -L"
     },
     "license": "BSD-3-Clause"
 }
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
         "eslint": "6.x.x",
         "find-rc": "4.x.x",
         "globby": "10.x.x",
-        "handlebars": "^4.7.3",
+        "handlebars": "4.x.x",
         "seedrandom": "3.x.x",
         "source-map": "0.7.x",
         "source-map-support": "0.5.x",
--- a/test/reporters.js
+++ b/test/reporters.js
@@ -1385,6 +1385,193 @@ describe('Reporter', () => {
     });
 
     describe('html', () => {
+        
+        it('generates corresponding lint-file parital when there are errors', () => {
+
+            const { lintFile } = require('../lib/reporters/html/partials/lint-file');
+
+            const options = {
+                errors: [
+                    {
+                        line: 19,
+                        severity: 'high',
+                        message: 'some message'
+                    }
+                ],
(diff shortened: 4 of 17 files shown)


package.json
```diff
@@ -44,7 +44,8 @@
     },
     "scripts": {
         "test": "node ./bin/_lab -f -L -t 100 -m 10000 --types-test test/index.ts",
-        "test-cov-html": "node ./bin/_lab -f -L -r html -m 10000 -o coverage.html"
+        "test-cov-html": "node ./bin/_lab -f -L -r html -m 10000 -o coverage.html",
+        "lint": "node ./bin/_lab -d -f -L --lint-fix"
```
aorinevo: Definitely remove `--lint-fix` if this gets removed. I recommend leaving lint script though.

test.html
```diff
@@ -0,0 +1,7 @@
+<script src="https://cdn.jsdelivr.net/npm/handlebars@latest/dist/handlebars.js"></script>
```
aorinevo: Remove this file.

README.md
```diff
@@ -15,3 +15,7 @@
 - [Changelog](https://hapi.dev/family/lab/changelog/)
 - [Project policies](https://hapi.dev/policies/)
 - [Free and commercial support options](https://hapi.dev/support/)
+
+
+# Notes
+- Replacing Handlebards with template strings.  Implement similar to https://medium.com/your-majesty-co/frameworkless-javascript-template-literals-the-best-thing-since-sliced-bread-d97f000ce955
```
aorinevo: Fix typo.

lib/reporters/index.js
```diff
@@ -81,7 +82,7 @@ exports.generate = function (options) {
         let output = '';
 
         reporter.report = function (text) {
-
+            // console.log('---- text: ', text);
```
aorinevo: remove console.log

test/reporters.js
```diff
@@ -1384,7 +1384,7 @@ describe('Reporter', () => {
         });
     });
 
-    describe('html', () => {
+    describe.only('html', () => {
```
aorinevo: Remove only.

aorinevo:
Status update:
- Fix 5 of 13 failing tests.
- Need to increase coverage on new `js` files.

aorinevo:
Status update:
- Only 2 tests left to fix.
- Still need to increase coverage for new `js` files.

README.md
```diff
@@ -15,3 +15,7 @@
 - [Changelog](https://hapi.dev/family/lab/changelog/)
 - [Project policies](https://hapi.dev/policies/)
 - [Free and commercial support options](https://hapi.dev/support/)
+
+
+# Notes
```
aorinevo: Remove

test/reporters.js
```diff
@@ -1427,7 +1614,7 @@ describe('Reporter', () => {
                 '<td class="source"><div>    while ( </div><div class="miss false" data-tooltip>value ) </div><div>{</div></td>']);
             expect(output, 'missed original line not included').to.contains([
                 '<tr id="while.js__14" class="miss">',
-                '<td class="source" data-tooltip>        value &#x3D; false;</td>']);
+                '<td class="source" data-tooltip>        value = false;</td>']);
```
aorinevo: This fails with feature flag off but passes with feature flag on.

It is possible to wrap this in the feature flag as well, but not sure it is worth the effort.

aorinevo:
All tests pass and coverage is at 100% with feature flag on.  There is one unit test that fails with feature flag off.  While it is possible to change the test to be backwards compatible, it might not be worth the effort (see  https://github.com/hapijs/lab/issues/977#issuecomment-605532859).

lib/reporters/html/partials/scripts.js
```diff
@@ -0,0 +1,89 @@
...
+
+      for (var i = 0, len = links.length; i < len; ++i) {
+          link = links[i];
+          link.className = link.getAttribute('href') == '#' + heading.id ? 'active' : '';
+      }
+  };
+
+  function find (y) {
+      var i = headings.length
+              , heading;
```
jugaltheshah: Not sure what's happening here, bit oddly formatted.

lib/reporters/html/partials/scripts.js
```diff
@@ -0,0 +1,89 @@
...
+
+      while (i--) {
```
jugaltheshah: I believe this would exit when `i === 0`, does headings[0] not need to be processed?

lib/reporters/html/partials/scripts.js
```diff
@@ -0,0 +1,89 @@
...
+              return heading;
+          }
+      }
+  }
+
+  function toggle (className) {
+
+      var elements = document.getElementsByClassName(className);
+
+      for (var i = 0, il = elements.length; i < il; ++i) {
```
jugaltheshah:
```suggestion
      for (var i = 0; i < elements.length; ++i) {
```
Same as before, I think this makes it a bit simpler.

lib/reporters/html/partials/scripts.js
```diff
@@ -0,0 +1,89 @@
...
+          }
+      };
+  }
+
+  function reset () {
+
+      var shownElements = document.getElementsByClassName('show');
+      var filterElements = document.querySelectorAll('input[type=checkbox]');
+
+      for (var i = 0, il = filterElements.length; i < il; ++i) {
```
jugaltheshah: Same as other loops, believe this can be simplified.

lib/reporters/html/partials/scripts.js
```diff
@@ -0,0 +1,89 @@
...
+          filterElements[i].checked = false;
+      }
+
+      // Check any filters with visible elements
+      for (i = 0, il = shownElements.length; i < il; ++i) {
```
jugaltheshah: Same here.

package.json
```diff
@@ -23,7 +23,7 @@
         "eslint": "6.x.x",
         "find-rc": "4.x.x",
         "globby": "10.x.x",
-        "handlebars": "4.x.x",
+        "handlebars": "^4.7.3",
```
jugaltheshah: Does this need to be pinned? If not, this would be better for consistency. 
```suggestion
        "handlebars": "4.x.x",
```

lib/reporters/html/partials/scripts.js
```diff
@@ -0,0 +1,89 @@
...
+        var current = document.querySelector('#menu a.active');
+        if (current) {
+          current.className = '';
+        }
+        return;
+      }
+      var links = document.querySelectorAll('#menu a')
+              , link;
+
+      for (var i = 0, len = links.length; i < len; ++i) {
```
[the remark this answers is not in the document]
aorinevo: Property lookup for links.length each and every time through the loop is wasteful, as the value doesn't change during the execution of the loop.  By storing the value on len var we avoid subsequent lookups.

lib/reporters/html/partials/scripts.js
aorinevo: See comment on first instance of this.

lib/reporters/html/partials/scripts.js
aorinevo: See comment on first instance of this.

lib/reporters/html/partials/scripts.js
aorinevo: See comment on first instance of this.

lib/reporters/html/partials/scripts.js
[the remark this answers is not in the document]
aorinevo: Ref: https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html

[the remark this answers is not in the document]
jugaltheshah: Array lookups are O(C) so either one should be equivalent. The advice in the link shared is specifically with regards to property lookups, which have the potential for additional costs due to needing to look up the prototype chain if local lookup fails.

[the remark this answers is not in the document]
jugaltheshah: Oops, I think I miunderstood. The idea is to make it a constant-time lookup rather than looking up the property on an Array object. The point I believe still stands though bc JS engines optimize for these sorts of things. E.g. see https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JIT_Optimization_Strategies.

lib/reporters/html/partials/scripts.js
DiegoRBaquero: 0 does get processed. The while Evals then the -1 is done

lib/reporters/html/partials/scripts.js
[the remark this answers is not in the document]
aorinevo: JIT compilation is _eventually_ performant.  First the JIT compiler needs to identify warm and hot code segments.  The warm code segments get sent to a baseline compiler and reused by the compiler where possible.  The hot code segments get sent to an optimizing compiler to be used until the assumptions under which the hot code was identified are invalidated at which point the optimized code is discarded (see https://blog.logrocket.com/how-javascript-works-optimizing-the-v8-compiler-for-efficiency/).

TL;DR there is a performance hit initially but indistinguishable for n sufficiently large, where n is the right hand of the for-loop conditional.

aorinevo:
Closing this PR for now as the vulnerability was addressed through a combination of PRs to Handlebars:
- wycats/handlebars.js#1666
- wycats/handlebars.js#1672

PR #987, opened 2020-08-11
throw error if an async experiment function is provided

timcosta:
Currently with code as follows:
```javascript
const Code = require('@hapi/code');
const Hoek = require('@hapi/hoek');
const Lab = require('@hapi/lab');

const { expect } = Code;
const { describe, it } = lab;

describe('Suite', async () => {
	
	await Hoek.wait(1000);

	it('should work', async () => {
		
		expect(1).to.exist();
	});
});
```
This test suite will be dropped and not reported silently due to the asynchronous `describe` function. If people are trying to do an async action within a describe block, they should use a `before` block with an async function.

This PR causes async experiment/describe/suite functions to throw an error instead of silently swallowing and hiding this type of issue.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -125,7 +125,11 @@ internals.experiment = function (title, options, fn) {
         this._setOnly(child, null, this._path);
     }
 
-    fn.call(null);  // eslint-disable-line no-useless-call
+    const returnValue = fn.call(null); // eslint-disable-line no-useless-call
+    // ensure that calls to experiment/describe are synchronous
+    if (typeof Hoek.reach(returnValue, 'then') === 'function') {
+        throw new Error(`Function for experiment '${title}' must be a synchronous function.`);
+    }
 
     this._titles.pop();
     this._path = this._titles.concat();                      // Clone
--- a/test/index.js
+++ b/test/index.js
@@ -108,6 +108,45 @@ describe('Lab', () => {
         expect(notebook.failures).to.equal(0);
     });
 
+    it('should throw on experiments with asynchronous functions', () => {
+
+        const script = Lab.script({ schedule: false });
+
+        expect(() => {
+
+            script.experiment('async function', async () => {
+
+                await Promise.resolve();
+            });
+        }).to.throw(/must be a synchronous function/);
+    });
+
+    it('should throw on experiments with asynchronous functions (BDD)', () => {
+
+        const script = Lab.script({ schedule: false });
+
+        expect(() => {
+
+            script.describe('async function', async () => {
+
+                await Promise.resolve();
+            });
+        }).to.throw(/must be a synchronous function/);
+    });
+
+    it('should throw on experiments with asynchronous functions (TDD)', () => {
+
+        const script = Lab.script({ schedule: false });
+
+        expect(() => {
+
+            script.suite('async function', async () => {
+
+                await Promise.resolve();
+            });
+        }).to.throw(/must be a synchronous function/);
+    });
+
     it('executes beforeEach and afterEach', async () => {
 
         let a = 0;


timcosta:
The check for whether an async function was passed is the only part I'm questioning. I went with detecting `then`-ables because that should cover functions that are defined as `async` and functions that return promises synchronously. If anyone has a better way to detect this, I'm all ears, but detecting a `then`-able was the safest route I could come up with today.

kanongil:
Shouldn't it just throw on any return value other than `undefined`?

timcosta:
I thought about asserting that the return value was falsey, but decided that would definitely be a breaking change as the documentation current doesn't state that you can't return anything. Happy to change the check to assert for an undefined return value though if that's the consensus.

nlf:
so the behavior as it is today is to simply not run the tests described in the block?

if so, i agree with this approach as a first step. this is less about keeping users from doing silly things that work by coincidence and more about making it more obvious why their tests are silently failing to run.

timcosta:
@nlf correct, the current behavior is that any tests in the async block are just not run at all, and they arent even shown in the test output as having been skipped or anything. it's as though the tests dont exist at all.

kanongil:
If we just want to handle the specific case, then I think it should check the function itself instead.

The test hierarchy setup is dependent on inline calls, and won't work if the tests are delayed to a future tick. Checking `fn.constructor.name === 'Function'` would also catch a similar error, if a generator with a `yield` is returned.

This will also make the error not runtime dependent. Example:

```js
describe('whatever', async function () {

    if (process.arch !== 'x64') {
        await 1;
    }

    it('sometimes goes here', () => {

        // …
    });
});

**Comment by nlf** [comment]
checking the constructor of the argument would catch a function defined as `async () => {}` or `async function () {}` but consider the case where someone is still using promises and writes

```js
describe('whatever', function () {
  return somethingAsync().then(function () {
    it('sometimes goes here', () => {
   })
  })
})
```

where the function is defined as a plain function, but the return value is a promise. we would encounter the same bug and receive no warning.

on top of that, if they were to not `return` in the above example we'd have no way to detect it at all since neither the function constructor check nor checking if the returned value is then-able would catch it.

i think the _real_ fix for this is to support async experiment functions but that's definitely non-trivial. in the mean time, i would bet that your initial example of `describe('whatever', async () => {})` is probably the most common failure pattern here and for that case checking the name of the constructor would be a good first step.

timcosta:
So are y'all requesting I change the implementation from catching any thenable to only catching certain types of thenables? Just want to make sure I understand, as checking the constructor name will only catch a subset of the issues that checking for a thenable will. No one has brought up a case that the current implementation would false positive on that is a realistic implementation, so I'm wondering why we would constrict the number of error cases we would catch.

nlf:
honestly? i'm pretty neutral. it's really 6 of one half a dozen of the other.

your proposed fix here would catch async functions as well as plain functions that return a promise, but it would not catch plain functions that run a promise but do not return them nor would it catch generators.

checking the name of the function constructor would catch async functions and generators, but not plain functions that return a promise or plain functions that run a promise but do not return it.

i don't think either one is more constricted than the other, we just have to make a call as to which failure cases are the most common and should be protected against.

timcosta:
We could do it as an `or`, so we check for thenables and generators. That wouldnt catch plan functions that run a promise but dont return it, but I'd probably consider that case out of scope.

nlf:
i think that's reasonable, that would catch 3 out of 4 possible failure scenarios with just 2 if statements

timcosta:
Updated. Opted for two `if`s in order to return the best error messages to the end user, and given that it isn't additional assertions.

PR #990, opened 2020-08-20
Use eslint-plugin-hapi embedded config

kanongil:
Draft changes needed to use the new plugin-based config from https://github.com/hapijs/eslint-plugin-hapi/pull/20. Final PR is dependent on it being published, and its name and version.

### DIFF
--- a/API.md
+++ b/API.md
@@ -583,9 +583,9 @@ As stated at the beginning of the document, `--ignore` parameter is an alias for
 
 **lab** uses a shareable [eslint](http://eslint.org/) config, and a plugin containing several **hapi** specific linting rules. If you want to extend the default linter you must:
 
-1. Add `@hapi/eslint-plugin-hapi` and `@hapi/eslint-config-hapi` as dependencies in your `package.json`. You must add both the plugin and the config because eslint treats them as peer dependencies. For more background, see [eslint/eslint#3458](https://github.com/eslint/eslint/issues/3458) and [eslint/eslint#2518](https://github.com/eslint/eslint/issues/2518).
+1. Add `@hapi/eslint-plugin-hapi` as a dependency in your `package.json`.
 
-2. In your project's eslint configuration, add `"extends": "@hapi/eslint-config-hapi"`.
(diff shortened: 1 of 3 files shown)


devinivy:
@cjihrig @kanongil are we in a place where eslint-plugin-hapi is ready to be published per https://github.com/hapijs/eslint-plugin-hapi/issues/21?

cjihrig:
I think so. Once https://github.com/hapijs/eslint-plugin-hapi/issues/21 happens, we need to publish a new version of `eslint-plugin-hapi`, but we also will want to deprecate `eslint-config-hapi` and the custom ESLint rules that now live in the plugin.

kanongil:
I have updated the draft PR with the renamed package. To finalise, it just needs an update to the entry in `package.json` with the actual release, after https://github.com/hapijs/eslint-plugin-hapi/pull/23 is merged.

package.json
```diff
@@ -15,7 +15,7 @@
     ],
     "dependencies": {
         "@hapi/bossy": "5.x.x",
-        "@hapi/eslint-config-hapi": "13.x.x",
+        "@hapi/eslint-plugin": "github:kanongil/eslint-plugin-hapi#rename",
         "@hapi/eslint-plugin-hapi": "4.x.x",
```
cjihrig, line 19: I published v5.0.0 of the eslint plugin with your changes.

kanongil: Awesome – I have finalized the PR.

API.md
```diff
@@ -583,9 +583,9 @@ As stated at the beginning of the document, `--ignore` parameter is an alias for
 
 **lab** uses a shareable [eslint](http://eslint.org/) config, and a plugin containing several **hapi** specific linting rules. If you want to extend the default linter you must:
```
Nargonath, line 584: Shouldn't we remove the mention of the shareable config since it is not directly exposed now?

kanongil: I updated the wording so it only reference the plugin (with a config).

Nargonath (approved):

cjihrig (approved):

Issue #989, opened 2020-08-20
Policy on updating typescript?


kanongil:
What is the policy for typescript updates? The integrated version (v3.6) is trailing behind, leaving out support for newer features like the `asserts` return type.

Since any of the frequent minor updates can break the tested types, would it make sense to change the dependency to a `peerDependency`? If combined with the [optional flag](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md), it can also help reduce the install size for non-typescript usage. Of course, this runs the risk of [a breaking change](https://github.com/microsoft/TypeScript/wiki/API-Breaking-Changes) to the [unstable API interface](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API) that lab uses.

The latter issue could possibly be remedied by doing a pre-validation of the expected interface used in [`types.js`](https://github.com/hapijs/lab/blob/master/lib/types.js), and throwing an error if it doesn't match. It could also be remedied, by creating a travis cron job, which tests against the current `latest`, `rc`, and `next` tags, and taking action if it fails.

cjihrig:
@kanongil since you added support for updating TS in https://github.com/hapijs/lab/pull/992, does this need to remain open?

kanongil:
Yes, this is fixed with #992 that delegates the responsibility away.

PR #992, opened 2020-08-21
Allow typescript to be updated by clients

kanongil:
This PR changes `typescript` to be a `peerDependency`, as suggested in #989.

As part of this, I have updated the integration and tests to handle typescript in the 3.6 - 4.1-pre range.

As a `peerDependency`, we won't have to create a new breaking release of lab for each minor update of the typescript compiler. We will only need to create a breaking release, if we need to stop supporting one or more old versions of the compiler. However, we will need to monitor the compatibility with the compiler, as new versions are released.

I decided to keep the existing version, as the minimally supported peer dependency for now.


geek:
@cjihrig fixed the milestone to be v24

kanongil:
I had another pass at this, and ended up revising the way that a missing typescript compiler is handled, by making the parent module (types.js), and other optional test pipeline components, into lazy loaded modules, handled by a new `lib/modules/index.js` loader.

This means that the approach is pretty hands-off, as lab will now not try to even load the `typescript` module, if not requested. If requested, eg. through `-Y`, it will throw a standard `Cannot find module 'typescript'` error, when it reaches that stage, which will be part of the output report.

This modularisation should should also make lab load faster, when not all features are in use.

kanongil:
The breaking change should probably also include the [eslint-plugin-hapi](https://github.com/hapijs/eslint-plugin-hapi/issues/21) renaming.

Nargonath (approved):

PR #993, opened 2020-09-01
Shallow clone context

### LINKED ISSUE #982: Error when using `context` to pass objects with private fields (e.g. hapi server object)
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: tested with `14.3.0` & `12.18.2`
* *module version with issue*: hapi `19.2.0` and lab `8.0.1`
* *last module version without issue*: hapi `18.4.1` and lab `???`
* *environment* (e.g. node, browser, native): `node`
* *used with* (e.g. hapi application, another framework, standalone, ...): using `@hapi/hapi` and `@hapi/lab`
#### What are you trying to achieve or the steps to reproduce?

[I originally posted this on Hapi's issue tracker.](https://github.com/hapijs/hapi/issues/4119) It was determined that the cause of this issue is using lab's `context` feature to pass the hapi `server` object to my tests. Minimal reproduction is available at [chrisshiplet/hapi-inject-bug-repro](https://github.com/chrisshiplet/hapi-inject-bug-repro) but here is the test in question:

```js
const Code = require("@hapi/code");
const Lab = require("@hapi/lab");

const getServer = require("../get-server");

const { expect } = Code;
const { it, describe, before } = (exports.lab = Lab.script());

describe("GET /", () => {
  before(async ({ context }) => {
    context.server = await getServer();

    await context.server.initialize();
  });

  it("responds with 200", async ({ context }) => {
    const res = await context.server.inject({
      method: "GET",
      url: `/some-path`,
    });

    expect(res.statusCode).to.equal(200);
  });
});
```

#### What was the result you got?

I get an error `Cannot read private member #strategies from an object whose class did not declare it`

#### What result did you expect?

I should be able to pass a server object around between tests. For now I've defined the context outside of my tests as mentioned in the Hapi issue, but that seems to defeat the point of having a context I can pass around in the first place. This is somewhat obscure and has an easy workaround, so I'm not even sure it's worth spending a lot of time on, but I wanted to post here for thoughts on it. Maybe the answer is just leaving a note in the context docs - right now it says `Tests aren't able to manipulate the context object for other tests.` but it wasn't clear to me what effect this object cloning would have on private fields, since they're a fairly new addition.

nlf:
This closes #982

### DIFF
--- a/test/runner.js
+++ b/test/runner.js
@@ -1816,71 +1816,90 @@ describe('Runner', () => {
 
         const script = Lab.script({ schedule: false });
 
-        const testContext = {
-            testData: { hello: 'there' }
-        };
-
-        const innerContext = {
-            testData: { goodbye: 'you' },
-            additionalData: { another: 'object' }
+        const contextData = {
+            testData: { hello: 'there' },
+            innerTestData: { goodbye: 'you' },
+            additionalData: { another: 'object' },
+            lateAddition: { more: 'data' }
         };
 
         script.experiment('test', () => {
 
+            let outerContext;
+
             script.before(({ context }) => {
 
-                context.testData = testContext.testData;
+                outerContext = context;
+                context.testData = contextData.testData;
             });
 
             script.test('has test context', ({ context }) => {
 
-                expect(context,'has proper context').to.equal(testContext);
-                expect(context,'is a shallow clone').to.not.shallow.equal(testContext);
-                expect(context.testData,'is a reference').to.shallow.equal(testContext.testData);
-                context.anotherProperty = { something: 'random' };
+                expect(context,'is deep equal').to.equal(outerContext);
+                expect(context,'is a shallow clone').to.not.shallow.equal(outerContext);
+                expect(context.testData,'is a reference').to.shallow.equal(contextData.testData);
+                context.additionalData = contextData.additionalData;
             });
 
             script.test('does not see changes to context from previous test', ({ context }) => {
 
-                expect(context,'has proper context').to.equal(testContext);
-                expect(context,'is a shallow clone').to.not.shallow.equal(testContext);
-                expect(context.testData,'is a reference').to.shallow.equal(testContext.testData);
-                expect(context.anotherProperty,'uses the original clone').to.not.exist();
+                expect(context,'is deep equal').to.equal(outerContext);
+                expect(context,'is a shallow clone').to.not.shallow.equal(outerContext);
+                expect(context.testData,'is a reference').to.shallow.equal(contextData.testData);
+                expect(context.additionalData,'ignores mutation from another test').to.not.exist();
             });
 
-            script.experiment('inner test', () => {
+            script.experiment('child experiment', () => {
+
+                let innerContext;
 
                 script.before(({ context }) => {
 
-                    context.testData = innerContext.testData;
-                    context.additionalData = innerContext.additionalData;
+                    innerContext = context;
+                    context.testData = contextData.innerTestData;
+                    context.additionalData = contextData.additionalData;
                 });
 
-                script.test('still has test context', ({ context }) => {
+                script.afterEach(({ context }) => {
 
-                    expect(context,'has proper context').to.equal(innerContext);
+                    context.lateAddition = contextData.lateAddition;
+                });
+
+                script.test('has the correct context', ({ context }) => {
+
+                    expect(innerContext,'is not the same reference').to.not.shallow.equal(outerContext);
+                    expect(context,'is deep equal').to.equal(innerContext);
                     expect(context,'is a shallow clone').to.not.shallow.equal(innerContext);
-                    expect(context.testData,'is a reference').to.shallow.equal(innerContext.testData);
-                    expect(context.additionalData,'is a reference').to.shallow.equal(innerContext.additionalData);
+                    expect(context.testData,'is a reference to inner data').to.shallow.equal(contextData.innerTestData);
+                    expect(context.additionalData,'is a reference').to.shallow.equal(contextData.additionalData);
+                    expect(outerContext.testData,'has not been changed').to.shallow.equal(contextData.testData);
                 });
 
-                script.after(({ context }) => {
+                script.test('receives context changes from afterEach', ({ context }) => {
 
-                    context.addedAfter = { another: 'object' };
+                    expect(innerContext,'is not the same reference').to.not.shallow.equal(outerContext);
+                    expect(context,'is deep equal').to.equal(innerContext);
+                    expect(context,'is a shallow clone').to.not.shallow.equal(innerContext);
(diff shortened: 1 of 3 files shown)


test/runner.js
```diff
@@ -1812,6 +1812,78 @@ describe('Runner', () => {
...
+
+            script.before(({ context }) => {
+
+                context.testData = testContext.testData;
+            });
+
+            script.test('has test context', ({ context }) => {
+
+                expect(context,'has proper context').to.equal(testContext);
+                expect(context,'is a shallow clone').to.not.shallow.equal(testContext);
```
devinivy, line 1838: I think to really test this you would have to use a separate variable to assign the `context` passed into before, then compare the test's `context` against that.  In short: even without shallow cloning this would pass.

test/runner.js
```diff
@@ -1812,6 +1812,78 @@ describe('Runner', () => {
...
+                expect(context.testData,'is a reference').to.shallow.equal(testContext.testData);
+                expect(context.anotherProperty,'uses the original clone').to.not.exist();
+            });
+
+            script.experiment('inner test', () => {
+
+                script.before(({ context }) => {
+
+                    context.testData = innerContext.testData;
```
devinivy, line 1855: I would be interested to see if the `context` passed here shallow equals (`===`) the one passed to the outer experiment's `before()` block.

test/runner.js
nlf: that assertion is only there to prove it's not a direct reference to the original testContext, i agree it's a bit weird but i wanted to make sure that was captured. if it's deep equal but not shallow equal, then it must be a shallow clone right?

test/runner.js
nlf: i was trying to keep assertions out of the before and after blocks since it seems to cause some weirdness.. do you not think the outer block's test immediately before this block is defined is enough?

devinivy: Hmm, good question.  What I'm picturing would look like this:
```js
            let beforeContext;
            script.before(({ context }) => {

                beforeContext = context;
                context.testData = testContext.testData;
            });

            // ...

                let innerBeforeContext;
                script.before(({ context }) => {

                    innerBeforeContext = context;
                    context.testData = innerContext.testData;
                });

                script.test('context is not shallow-cloned across before()s', ({ context }) => {

                    expect(beforeContext,'is a reference').to.shallow.equal(innerBeforeContext);
                });
```

kanongil (commented):
Why change this? The feature is esoteric, and not used inside the hapijs org (outside of testing the feature in lab itself).

I say, if we want to do a breaking change, just remove the feature entirely.

nlf:
why remove it entirely? there's some value in discouraging mutation of the context in tests to prevent one test relying on another test having passed.

this change is intended to allow you to do things like attach a server or a database client to the context in your before/after functions, which is a common pattern and does not work today.

test/runner.js
nlf: ahhhh yes, ok i see what you're saying. i'll tweak things to compare the context object in each location directly and remove the nonsense equality assertions that are here now

devinivy:
I'm personally for keeping it with this fix, though I don't have especially strong feelings.  Aside from encouraging good hygiene, it's also sort of convenient over initializing variables then assigning them in `before()`.

kanongil:
Common pattern where? It is not used in any hapijs org repos.

When you remove the cross test deep clone part of the feature, what remains is a shorthand to defining variables in the before() scope.

```js
describe('whatever', () => {

    before(async ({ context }) => {

        context.server = await startServer();
    });

    it('does something with a server', ({ context: { server } }) => {

        expect(server.started).to.be.true();
    });
});
```

vs.

```js
describe('whatever', () => {

    let server;

    before(async () => {

        server = await startServer();
    });

    it('does something with a server', () => {

        expect(server.started).to.be.true();
    });
});
```

While it can save a line of code, it adds an extra argument to tests, and loses any possibility of type inference for eg. autocompletion.

Also, this is not a bugfix, but a breaking change.

devinivy:
For what it's worth, it is recognized that this is a breaking change— it would go into v24 with #992.  I think that slight aesthetic difference between your two examples is exactly what people intend to get out of the `context` feature (outside the hapi org).  When I said "common" I meant that it's common to share some resources across tests, as shown in your second example.

nlf:
as the commit messages state, this is known to be a breaking change and not a bug fix. i added the breaking changes label to it to make that even more clear.

as @devinivy mentions, your example clearly illustrates how people _expect_ it to work, which is not how it works today. yes, it is a convenience change and historically would've been rejected since it is not necessary for how it is consumed by the hapijs org today, but we live in a brave new world where useful changes can be accepted regardless of whether or not the hapijs core uses them.

i'm not sure i follow your concern about type inference, a shallow clone and a deep clone would be the same inferred type would they not? the existing behavior already allowed the `context` to be a different type across experiments and that hasn't changed.

devinivy:
Cool, let's figure this out.  It seems like we have three options:
 - close this PR and release v24 without any changes to this feature.
 - wait to release v24 until we come to consensus on this question and possibly do some additional development.
 - release this work in v24 and continue the conversation.

Given that this work is already complete and seems to be strictly an improvement to an already-existing feature (with an open issue #982), I propose we release it in v24 but create an issue to discuss and come to consensus whether the `context` feature should be removed altogether.  How does this sound to everyone?  Gotta love meta-consensus :)

Nargonath (approved):
I'm in favor of keeping the `context` feature for now. Personally I don't use it much but there seems to be people who do. I think it is a good idea to open an issue afterwards and discuss it to see if there is an interest in deleting the feature altogether if we really want to go that route.

kanongil:
FYI, I'm fine with changing it. It just seems to make the feature less valuable.

I will need to change a couple of tests that rely on the cloning behaviour, where I will replace it with a solution without the `context` object, and thus wondered whether it even makes sense to keep around.

kanongil:
> i'm not sure i follow your concern about type inference, a shallow clone and a deep clone would be the same inferred type would they not? the existing behavior already allowed the `context` to be a different type across experiments and that hasn't changed.

The type inference argument does not apply to this change, but in general to attaching stuff to the `context` object, vs. a scoped variable.

devinivy:
> I will need to change a couple of tests that rely on the cloning behaviour

I think this would be useful to see, since currently nobody has shown a concrete use-case for the deep cloning behavior, but we have seen its drawbacks (e.g. you can't place a hapi server on `context`).  Could you give a brief example?

kanongil:
Sure, the code is [here](https://github.com/kanongil/node-m3u8parse/blob/b8955f3e266c9e93e9e3d15f32421ec18eef45cb/test/attrlist.test.js#L124-L127).

Of course, it won't actually break with this change, since I only modify the object that one place.

kanongil (approved):

geek (approved):

PR #1000, opened 2021-01-06
Fix retrying of passing tests

kanongil:
Any test configured with a `retry` more than 1 will erroneously re-run the test logic until all tries are used.

The issue is fixed with this PR.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -654,6 +654,7 @@ internals.protect = function (item, state) {
 
             item.tries = 0;
             let retries = !item.options.retry ? 1 : (item.options.retry === true ? state.options.retries : item.options.retry);
+            let lastError;
             while (retries--) {
                 ++item.tries;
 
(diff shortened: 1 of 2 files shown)


devinivy (approved):
Well, that was a funny bug!  Thanks for catching and fixing it.

cjihrig (approved):
😄

kanongil:
Funny, yeah - This is a class of bug that tests likely won't catch. The code appears to run as expected, but is just a bit slow. At least it was simple enough to write a failing test.

Though in this case, it was not only slow. Given that the test code would be run after the teardown in `finish()`, I expect that it could be the cause of some weird issues. In this case I discovered it since all the catbox policy tests were reporting 2 tries, when I would only ever expect 1, and it in fact used 5.

lloydbenson:
This was a great find. Thanks @kanongil.

kanongil:
Can you do a patch release?

devinivy:
@kanongil consider it done 👍

PR #1001, opened 2021-03-04
Introduce an upper limit to how much actual vs. expected is diffed

rluba:
When (accidentally) asserting on very big object graphs (like Node.js’ response objects), the actual / expected strings generated by the `console` logger can become hundreds of MB long.

In this case the reporter stalls for a very, very long time while trying to create diffs for these huge strings. Adding an upper limit reduces those stalls to a bearable amount while still remaining useful for most conceivable applications.

### DIFF
--- a/lib/reporters/console.js
+++ b/lib/reporters/console.js
@@ -209,8 +209,18 @@ internals.Reporter.prototype.end = function (notebook) {
             if (test.err.actual !== undefined &&
                 test.err.expected !== undefined) {
 
-                const actual = internals.stringify(test.err.actual);
-                const expected = internals.stringify(test.err.expected);
+                let actual = internals.stringify(test.err.actual);
+                let expected = internals.stringify(test.err.expected);
+
+                if (this.settings['max-diff-length'] > 0) {
+                    const maxLen = this.settings['max-diff-length'];
+                    if (actual.length > maxLen) {
+                        actual = actual.substr(0, maxLen) + '[truncated]';
+                    }
+                    if (expected.length > maxLen) {
+                        expected = expected.substr(0, maxLen) + '[truncated]';
+                    }
+                }
 
                 output += '      ' + whiteRedBg('actual') + ' ' + blackGreenBg('expected') + '\n\n      ';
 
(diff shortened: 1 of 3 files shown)


rluba:
Checks seem to fail due to two uncovered lines. But these lines were are already unconvered in `master` – the total was just reported as "100%" due to rounding errors. (See `coverage.html` after testing `master`.)

kanongil:
Makes sense with a truncation.

Don't know if I am a fan of a new option. Personally, I would just choose a sane value and hardcode it.

rluba:
Yeah, I was thinking about that tradeoff as well. I just wanted to give people with weird edge-cases a chance to overwrite our sane value, if they need to.

lib/reporters/console.js
```diff
@@ -209,8 +209,19 @@ internals.Reporter.prototype.end = function (notebook) {
...
+
+                if (this.settings['max-diff-length'] > 0) {
+                    const maxLen = this.settings['max-diff-length'];
+                    if (actual.length > maxLen) {
+                        actual = actual.substr(0, maxLen) + '[truncated]';
+                    }
+
+                    if (expected.length > maxLen) {
+                        expected = expected.substr(0, maxLen) + '[truncated]';
+                    }
```
Nargonath, line 223: Shouldn't we test for `(actual.length + expected.length) > maxLen`? In the end with the current code you can print `2 * 20 * 2013` to the console. Isn't it too much or did you already took that into account?

Nargonath (commented):
I'd be also in favor of KISS and just hardcode the value. We can always reassess later if an issue is raised for it.

devinivy:
I agree with hardcoding the value for now as well.  If we make that simplification and address @Nargonath's review, I think this will be good to go!

lib/reporters/console.js
rluba: >  Isn't it too much or did you already took that into account?

The limit is mostly about not overwhelming the diff algorithm (as it gets very slow for larger strings). Since both strings are diff’d separately, the limit can be applied separately.

Since this is not a hard limit like a buffer size, I don’t think it makes any sense to add any complexity to take the suffix into account.

rluba:
The failed check seems to be a pre-existing flaky test. (It also failed locally from time to time.)

hueniverse:
The fix above applies the changes requested (replace the config with a hard-coded value).

PR #1005, opened 2021-04-21
removing error truncation from types module

saboya:
This helps when debugging error messages if they´re too long.

### DIFF
--- a/lib/modules/types.js
+++ b/lib/modules/types.js
@@ -14,6 +14,7 @@ const Utils = require('../utils');
 
 const internals = {
     compiler: {
+        noErrorTruncation: true,
         strict: true,
         jsx: Ts.JsxEmit.React,
         lib: ['lib.es2020.d.ts'],


Nargonath (commented):
This LGTM but on the TypeScript documentation they mention that this option is deprecated: https://www.typescriptlang.org/tsconfig#noErrorTruncation. I'm not really sure what to think about that since they don't mention any other replacement for it. There would be no point in adding an option bound to be removed in future versions IMO.

saboya:
I see. I searched for anything resembling a discussion about that in the typescript repo and found nothing, which is pretty weird. Not sure what ot think about it too.

Nargonath:
I'll ask in our TypeScript slack channel just in case someone knows better.

Nargonath:
FWIW I couldn't not find any explanation for the deprecation in the TS repo either. I wonder if it's not a documentation error. 🤔

kanongil:
Hmm, the deprecation was introduced in the docs with https://github.com/microsoft/TypeScript-Website/commit/d65e30fe5473656bc34da45f4454259ee29224e7. It can now be found [here](https://github.com/microsoft/TypeScript-Website/blob/e12b9b5e2a533331a4765d9468ae99b47fad1163/packages/tsconfig-reference/scripts/tsconfigRules.ts#L26), listed under *Things we should document, but really want to help move people away from*.

I think that is about all the reasoning you are going to get.

saboya:
Not sure how they are going to move people away from an option that has no alternative =P

Nargonath:
Thanks @kanongil, I failed to find the information. 🙂

I believe that's something you'd need only in specific cases perhaps that's the reasoning behind it. I don't mind if we choose to add it to be honest. Let's see what other think about it. It will just make it more verbose for every TS users. Perhaps that's something we could add behind a command-line flag (not entirely sure though, I didn't dig through it).

saboya:
Hiding it behind a command-line flag sounds reasonable, I can do that if we decide to move foward with it.

PR #1013, opened 2021-06-20
Add eslint-plugin module config to lint/coverage feature

Nargonath:
Since we published a new version of the `@hapi/eslint-plugin` with a new configuration `@hapi/module` we want to migrate the inner configuration declared in `lib/linter/.eslintrc.js` to use the new configuration.

### DIFF
--- a/lib/modules/coverage.js
+++ b/lib/modules/coverage.js
@@ -18,7 +18,7 @@ const Transform = require('./transform');
 const internals = {
     ext: Symbol.for('@hapi/lab/coverage/initialize'),
     _state: Symbol.for('@hapi/lab/coverage/_state'),
-    EslintEngine: new ESLint.CLIEngine({ baseConfig: Eslintrc }),
+    EslintEngine: new ESLint.CLIEngine({ baseConfig: Eslintrc })
 };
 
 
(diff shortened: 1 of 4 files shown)


lib/modules/coverage.js
```diff
@@ -330,7 +331,13 @@ internals.instrument = function (filename) {
 
     // Parse tree
 
-    const tree = BabelESLint.parse(content, Eslintrc.parserOptions);
+    const eslintConfig = internals.EslintEngine.getConfigForFile(filename);
```
Nargonath, line 334: There is an error with the coverage feature regarding the `lib/linter/.eslintrc.js` file. You can find the error in the CI log: https://github.com/hapijs/lab/pull/1013/checks?check_run_id=2868449419. I can't seem to find a solution. FWIU the problem comes from this line but it seems to work properly for the rest of the `lab` codebase but not for the `.eslintrc.js` file. Perhaps I miss configured `eslint`. If anyone has a clue here, I'm all up for it.

Nargonath: @nlf found the root of the problem 🙏 :
image

nlf (approved):
this seems reasonable to me, and we know how to fix it if someone is mucking about in the require cache and it becomes a problem. i'm a big fan of starting simple 👍

Nargonath:
I totally agree. Thanks. 👍

PR #1015, opened 2021-07-12
fix: correct semver range for @hapi/eslint-plugin

nlf:
lab tries to extend a config that doesn't exist in 5.0.0 of eslint-plugin, so the `5.x.x` semver here is wrong. switched it to the standard caret syntax so we can express what we actually need, which is _at least_ 5.1.0


Nargonath (approved):
Good catch, thank you @nlf . There are failing tests but I believe those are just flacky tests. I'll rerun the jobs.

Issue #972, opened 2020-02-25
Original files in HTML reporter


gerardolima:
#### Support plan

* *which support plan is this issue covered by?* Community
* *is this issue currently blocking your project?* no
* *is this issue affecting a production system?* yes

#### Context

* *node version*: v12.13.1
* *module version with issue*: 22.0.3
* *last module version without issue*: unknown
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi application
* *any other relevant information*: this issue _should_ be fixed by #648; it is related to #614, #338 and #952. 

#### What are you trying to achieve or the steps to reproduce?

command
```bash
lab --coverage --sourcemaps --reporter html --output tmp/coverage.html src; open tmp/coverage.html;
```
_.labrc.js_
```javascript
module.exports = {
	// file-macthing
	flat: false,
	paths: ['./src'],
	pattern: 'spec',
	transform: 'node_modules/lab-transform-typescript',
	assert: false,
	colors: true,
	coverage: false,
	globals: '__core-js_shared__,core',
	lint: false,
	timeout: 4000,
	verbose: true,
	sourcemaps: true
};
```

#### What was the result you got?
The coverage report refers to the transformed file (in Javascript).

#### What result did you expect?
I expected the coverage report to refer to the source files (in Typescript).

hueniverse:
I am more inclined to remove sourcemaps support completely then try to fix this. There are other test framework that work much better for typescript. This module is just not going to solve all this problems as well.

gerardolima:
hey, @hueniverse, I'm not sure I understood what you meant with _This module is just not going to solve all this problems_. You mean the html reporter? `@hapi/lab`? or `lab-transform-typescript`?

PR #1029, opened 2022-01-18
Add inline typescript sources to transformed output

kanongil:
The html reporter cannot resolve the path to the original typescript source files which means that it only displays the transformed code. This PR fixes the issue by including the source in the inline source-map.


kanongil:
This should actually fix #972.

devinivy (approved):
Thanks!

PR #1031, opened 2022-02-14
Support for ES modules

### LINKED ISSUE #1027: Lab is not working with import syntax in my project
#### Support plan

* *is this issue currently blocking your project?* (yes/no): yes
* *is this issue affecting a production system?* (yes/no): no

#### Context

* *node version*: 14.17.6
* *module version*: 24.5.0
* *environment* (e.g. node, browser, native): node API
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi 
* *any other relevant information*: I set "type": "module" in package.json

#### How can we help?

Hello,

I have an issue with Lab, I can't perform tests with my current project.
All my app use "import" syntax with type module activated in package.json. I don't have Babel and I don't use Typescript.
It's a very simple project for now.

Here the error when I try to launch my npm script :

```
Error requiring file: D:\Projects\new-api\test\server.test.js
Must use import to load ES Module: D:\Projects\new-api\test\server.test.js
require() of ES modules is not supported.
require() of test\server.test.js from  D:\Projects\new-api\node_modules\node_modules\@hapi\lab\lib\cli.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: D:\Projects\new-api\routes\server.test.js
require() of ES modules is not supported.
require() of D:\Projects\new-api\routes\server.test.js from D:\Projects\new-api\node_modules\@hapi\lab\lib\cli.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename server.test.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from D:\Projects\new-api\package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1089:13)
    at Module.load (internal/modules/cjs/loader.js:937:32)
    at Function.Module._load (internal/modules/cjs/loader.js:778:12)
    at Module.require (internal/modules/cjs/loader.js:961:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at D:\Projects\new-api\node_modules\@hapi\lab\lib\cli.js:135:13
    at Array.forEach (<anonymous>)
    at Object.internals.traverse (D:\Projects\new-api\node_modules\@hapi\lab\lib\cli.js:129:15)
    at Object.exports.run (D:\Projects\new-api\node_modules\@hapi\lab\lib\cli.js:60:31)
    at main (D:\Projects\new-api\node_modules\@hapi\lab\bin\lab:56:50)
    at Object.<anonymous> (D:\Projects\new-api\node_modules\@hapi\lab\bin\lab:60:1)
    at Module._compile (internal/modules/cjs/loader.js:1072:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
    at Module.load (internal/modules/cjs/loader.js:937:32)
    at Function.Module._load (internal/modules/cjs/loader.js:778:12)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)
    at internal/main/run_main_module.js:17:47
```

Here server.test.js code :

```js
import * as Lab from '@hapi/lab';
import { expect } from '@hapi/code';
import { init } from '../index';

const lab = Lab.script();
const { describe, it, beforeEach,  afterEach} = lab;

describe('GET /', () => {
    let server;

    beforeEach(async () => {
        server = await init();
    });

    afterEach(async () => {
        await server.stop();
    });

    it('responds with 200', async () => {
        const res = await server.inject({
            method: 'get',
            url: '/'
        });
        expect(res.statusCode).to.equal(200);
    });
});

```
And my npm script:
```bash
"test": "lab ./routes/server.test.js"
```
It is confusing because I've tried a lot of things and nothing is working and I don't have this problem with another package.
Am I missing something ?

devinivy:
This implements basic ES module support in lab.  There are some technical blockers to implementing granular, branch-level code coverage for ESM, so I punted on that for now by a. not allowing users to use coverage options with ESM test files, and b. directing users to documentation on how to use [c8](https://github.com/bcoe/c8) with lab, which at least gives users a decent coverage option for now.

I expect there will be some finer points to tighten-up through code review and perhaps some real-life testing, but I wanted to make this available for feedback since I think it is largely complete.

From the docs, here are the current blockers with coverage:
> Lab does not support code coverage for ES modules.  There are two reasons for this: in order to implement this we would either use [V8's builtin coverage](https://v8.dev/blog/javascript-code-coverage) or an [ESM Loader](https://nodejs.org/api/esm.html#loaders).  Unfortunately the former [doesn't support](https://bugs.chromium.org/p/v8/issues/detail?id=10627) granular branch coverage as we do in lab, and the latter is an experimental API that is still settling.  We hope to provide ESM coverage support in the future once one or both of these issues are resolved.

Resolves #1027

### DIFF
--- a/API.md
+++ b/API.md
@@ -6,7 +6,7 @@
 
 ## Usage
 
-By default, **lab** loads all the '\*.js' files inside the local 'test' directory and executes the tests found. To use different directories or files, pass the file or directories as arguments:
+By default, **lab** loads all the '\*.js|cjs|mjs' files inside the local 'test' directory and executes the tests found. To use different directories or files, pass the file or directories as arguments:
 
 ```bash
 $ lab unit.js
@@ -641,7 +641,7 @@ lab -dL
 
 ## Integration with an assertion library
 
-Using the `--assert` argument allows you to integrate Lab with your favorite assertion library. Aside from `--assert` from the CLI you can change the `assert` option when executing `report`. Whatever assertion library you specify is imported and assigned to the `Lab.assertions` property. Here is an example using `lab --assert code`:
+Using the `--assert` argument allows you to integrate Lab with your favorite assertion library. Aside from `--assert` from the CLI you can change the `assert` option when executing `report`. Whatever assertion library you specify is imported and assigned to the `Lab.assertions` property. Here is an example using `lab --assert @hapi/code`:
 
 ```js
 const lab = exports.lab = Lab.script();
@@ -665,7 +665,7 @@ describe('expectation', () => {
 ```
 
 ```
-$ lab --assert code
+$ lab --assert @hapi/code
 ```
 
 If you use the [Code](https://github.com/hapijs/code) assertion library Lab will let you know if you have any missing assertions. An example of this is:
@@ -736,6 +736,27 @@ See the [json reporter](lib/reporters/json.js) for a good starting point.
 
 ## Coverage
 
+### ESM support
+
+Lab does not support code coverage for ES modules.  There are two reasons for this: in order to implement this we would either use [V8's builtin coverage](https://v8.dev/blog/javascript-code-coverage) or an [ESM Loader](https://nodejs.org/api/esm.html#loaders).  Unfortunately the former [doesn't support](https://bugs.chromium.org/p/v8/issues/detail?id=10627) granular branch coverage as we do in lab, and the latter is an experimental API that is still settling.  We hope to provide ESM coverage support in the future once one or both of these issues are resolved.
+
+In the meantime, we recomming using lab with [c8](https://github.com/bcoe/c8) in order to provide code coverage in ESM projects.  Note that c8 does not support granular branch coverage the way we do in lab, for the same reasons listed above.  It's pretty simple to use c8 with lab, though.
+
+First install c8:
+```
+npm install --save-dev c8
+```
+
+Next update your test command to start with c8. Here's an example in a package.json file switching from lab's coverage to c8's coverage:
+```diff
+   "scripts": {
+-      "test": "lab -a @hapi/code -t 100"
++      "test": "c8 --100 lab -a @hapi/code"
+   },
+```
+
+### Inline enabling/disabling
+
 Sometimes you want to disable code coverage for specific lines, and have the coverage report omit them entirely. To do so, use the `$lab:coverage:(off|on)$` comments. For example:
 ```javascript
 // There is no way to cover this in node 0.10
@@ -746,7 +767,7 @@ if (typeof value === 'symbol') {
 /* $lab:coverage:on$ */
 ```
 
-### Coverage Bypass Stack
+### Coverage bypass stack
 
 Disabling code coverage becomes tricky when dealing with machine-generated or machine-altered code. For example, `babel` can be configured to disable coverage for generated code using the `auxiliaryCommentBefore` and `auxiliaryCommentAfter` options. The naïve approach to this uses `$lab:coverage:on$` and `$lab:coverage:off$`, but these directives overwrite any user-specified directives, so that a block where coverage should be disabled may have that coverage unintentionally re-enabled. To work around this issue, `lab` supports pushing the current code coverage bypass state onto an internal stack using the `$lab:coverage:push$` directive, and supports restoring the top of the stack using the `$lab:coverage:pop$` directive:
 ```javascript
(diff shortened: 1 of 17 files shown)


lib/cli.js
```diff
@@ -57,22 +58,29 @@ exports.run = function () {
         require('source-map-support').install(sourceMapOptions);
     }
 
-    const scripts = internals.traverse(settings.paths, settings);
+    const scripts = await internals.traverse(settings.paths, settings);
+
+    const isNotCjs = (script) => script._type !== 'cjs';
+    if (settings.coverage && scripts.some(isNotCjs)) {
```
devinivy, line 64: Perhaps we should have similar checks for `settings.typescript` and `settings.transform`.  I am not sure how common it could be to find projects mixing both ESM and typescript/babel.  It could definitely lead to some confusing issues.  If we'd like to cater to it, another option is to run the tests but output a warning and exit `1` unless it is acknowledged by passing some kind of additional flag.

geek (approved):

lib/cli.js
kanongil: Hmm, yeah transforms, including typescript, also won't work through `import`. It would make sense to reject those as well.

While it is unlikely to combine ESM modules with imports from typescript code, it could happen while migrating. FYI, I already have projects which has commonjs test cases, but the rest of the logic is based on typescript.

Typescript has experimental ESM node.js output support to complicate matters further: https://www.typescriptlang.org/docs/handbook/esm-node.html

lib/runner.js
```diff
@@ -139,7 +139,7 @@ exports.execute = async function (scripts, options, reporter) {
 
     const experiments = scripts.map((script) => {
 
-        script._executed = true;
+        script._executionScheduled = true;
         return script._root;
     });
```
kanongil, line 144: Isn't this redundant now?

kanongil (commented):
This seems like a fine first step to support modules, with limited changes.

I haven't reviewed the technical details, so this is only some loose comments.

We should probably have a look at our eslint config to better support the new module syntax and maybe append the styleguide. This should probably happen before something like this is merged, since subsequent changes could cause the linter to fail for existing code.

API.md
```diff
@@ -736,6 +736,27 @@ See the [json reporter](lib/reporters/json.js) for a good starting point.
 
 ## Coverage
 
+### ESM support
+
+Lab does not support code coverage for ES modules.  There are two reasons for this: in order to implement this we would either use [V8's builtin coverage](https://v8.dev/blog/javascript-code-coverage) or an [ESM Loader](https://nodejs.org/api/esm.html#loaders).  Unfortunately the former [doesn't support](https://bugs.chromium.org/p/v8/issues/detail?id=10627) granular branch coverage as we do in lab, and the latter is an experimental API that is still settling.  We hope to provide ESM coverage support in the future once one or both of these issues are resolved.
+
+In the meantime, we recomming using lab with [c8](https://github.com/bcoe/c8) in order to provide code coverage in ESM projects.  Note that c8 does not support granular branch coverage the way we do in lab, for the same reasons listed above.  It's pretty simple to use c8 with lab, though.
```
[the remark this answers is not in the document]
devinivy: I will add a note about that 👍

lib/runner.js
devinivy: It's a fair question.  When you invoke scripts with the lab CLI it is definitely redundant since `_executionScheduled` is already marked `true`.  If the script is self-invoked (e.g. `node test/my-test-script.js`) or by using lab as a module then it will be set from `false` to `true`, though it may never be read by lab after that anyway.  Overall I lean towards keeping it for consistency in those latter two situations.

kanongil: But it hasn't been scheduled, it's actively running at this point. `_executionScheduled` just guards the automatic execution when the test code calls `.script({ schedule: true })`, and should not need to be set otherwise.

Nargonath (approved):

Nargonath:
Well done @devinivy. 👍

lib/runner.js
devinivy: I took a look at this, and it turns out that it's necessary to keep this in order for the test suite to function properly, due to the way scripts are sometimes executed directly within the tests.  I assume that is why this check was here to begin with.  It's just an "internal" concept, but I don't mind the idea that every script that's executed is also marked as having been scheduled, even if it was run directly.

devinivy:
I've added a note about the coverage comment incompatibility with c8.  I have also done a little dive into our lint rules, and they do seem to be fairly compatible with ESM.  After speaking with @geek and @Nargonath, we feel comfortable shipping this functionality without any changes to our lint rules for the time being.  We are happy to reconsider and ship a batch of breaking changes if folks find rough edges with the lint rules and ESM.  Also, these should not affect hapi itself since hapi will remain written in CJS for the foreseeable future.  There should be a new major version in the near future to drop node v12 support and support eslint v8.

PR #1036, opened 2022-04-10
Coverage for nullish coalescing operator

devinivy:
Supporting coverage for nullish coalescing operator `??`.  Previously it would flag false positives for coverage issues.


kanongil (approved):
Awesome, but is this not something that should be merged into the current release line? We claim to support node versions that can use this operator.

FYI, I added a test to verify that lab can detect missing coverage on both sides.

devinivy:
Yeah, you're right, it should.  It will just be a bit of a rigamarole to bypass coverage for nullish coalescing on node v12, but it's worth it.  On it!

PR #1037, opened 2022-04-10
Remove undocumented --coverage-module option

kanongil:
I'm not aware of any existing usage of this half-baked feature, and it seems like a good candidate for inclusion in the v25 release.

This also fixes `npm install` on the lab repo, so it doesn't complain about the moved `@hapi/pinpoint` dependency, and `@hapi/lab-external-module-test` can now be deprecated.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -229,12 +229,6 @@ internals.options = function () {
             description: 'prevent recursive inclusion of all files in coveragePath in report',
             default: null
         },
-        'coverage-module': {
-            type: 'string',
-            description: 'enable coverage on external module',
-            multiple: true,
-            default: null
-        },
         'coverage-path': {
             type: 'string',
             description: 'set code coverage path',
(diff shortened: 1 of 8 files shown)


devinivy:
We could probably use additional opinions on this.  I understand that it is currently undocumented, but I also think this functionality of lab is pretty novel/special, and is compatible with coverage of joi schemas: https://github.com/sideway/joi/blob/95feacc295bc1d72412c128c76f8b1c989576c1e/lib/trace.js#L23-L38

Here's a little demo I have created to show how this works with joi: https://github.com/devinivy/joi-coverage-example

kanongil:
Thanks for enlightening me on how this is used, which helped me understand what it does. Enabling joi to validate the rules is very nice, but this seems like a complicated way to add a custom coverage report.

I might be wrong, but couldn't this be better served using a [custom reporter](https://hapi.dev/module/lab/api/?v=24.5.1#custom-reporters)? Eg. as `lab -c -r console -r joi/coverage`. This also means that there is no weird discrepancy, where the result is only reported using the console and json reporters (other reporters will just succes, but lab will return with an error code).

kanongil:
So I looked into the practicality of doing it as a custom reporter, and it seems to work quite nicely. The main issue is that there is no nice way to make a reporter trigger a failure. See my patch here: https://github.com/kanongil/joi-coverage-example/commit/1ae4b4ae727fadc3a787e8c1cb965c0b1ee6cea3.

For joi it would probably work best as a separate module, along with some of the tracing logic from `lib/trace.js`.

Let me know what you think?

Nargonath (approved):
I wasn't aware of this feature. According to `git blame`, it's been in `lab` for 3 years now and nobody raised an issue since regarding the lack of documentation so I'd expect most people to not use it. Since it was added by @hueniverse he might have a personal usage of this feature. Would it be a problem for you if we remove it from lab @hueniverse?

I like @kanongil's suggestion. We don't need to burden lab's codebase with a feature nobody's using but there is still a way to use it as an external dep if people feel like it.

Nargonath:
I like the idea also to be honest. I'm not sure we really want it to be removed. Do you think it's not worth keeping the feature if we enable it through a custom reporter as Gil suggested?

devinivy:
I always thought of reporters as being to control the presentation of the coverage report (i.e. should it be written to a file as HTML? or to stdout as lcov?  both?), but if it can also provide comparable functionality to coverage-module then that's great.

devinivy:
Fixed the lint issue affecting the CI run prior to merge: https://github.com/hapijs/lab/commit/481676b2486f4889aae2418d2c55830518e4a01f

PR #1039, opened 2022-04-10
Update diff package

kanongil:
Targeting this at v25, since the release appears to [change the output slightly](https://github.com/kpdecker/jsdiff/blob/master/release-notes.md#v500).


devinivy (approved):
Thanks!

Issue #1033, opened 2022-04-02
tsconfig paths alias


sahitono:
#### Support plan

* *is this issue currently blocking your project?* (no):
* *is this issue affecting a production system?* (no):

#### Context

* *node version*: 16.13.2
* *module version*:  24.5.1
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi application
* *any other relevant information*: used in typescript project

#### How can we help?

Hi, I am currently using tsconfig paths alias like this
```json
        "paths": {
            "~/*": [
                "./*"
            ]
        },
```
but it looks the built in typescript transformer doesn't support tsconfig paths alias yet, Is there any way to integrate the transformer it with tsconfig-paths or tsc-alias?

thank you in advance

devinivy:
I would personally be open to bringing this functionality into lab, and would be happy to review someone's work on this.

devinivy:
This was resolved by #1041, which was published in 24.7.0.

PR #1041, opened 2022-04-15
feat: 🎸 depedencies cli option

damusix:
Addresses https://github.com/hapijs/lab/issues/1033

Uses `tsconfig-paths/register` as test.

> Needs documentation

Usage: 
```
lab --typescript --require 'tsconfig-paths/register' test
```

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -252,13 +252,6 @@ internals.options = function () {
             description: 'minimum plan threshold to apply to all tests that don\'t define any plan',
             default: null
         },
-        require: {
-            alias: 'req',
-            multiple: true,
-            type: 'string',
-            description: 'dependency module to require and run before tests run (useful for things like babel, tsconfig-paths, etc)',
-            default: null
-        },
         dry: {
             alias: 'd',
             type: 'boolean',
@@ -364,6 +357,13 @@ internals.options = function () {
             multiple: true,
             default: null
         },
+        require: {
+            alias: 'req',
+            multiple: true,
+            type: 'string',
+            description: 'dependency module to require and run before tests run (useful for things like babel, tsconfig-paths, etc)',
+            default: null
+        },
         retries: {
             alias: 'R',
             type: 'number',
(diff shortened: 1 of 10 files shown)


lib/cli.js
```diff
@@ -252,6 +252,12 @@ internals.options = function () {
             description: 'minimum plan threshold to apply to all tests that don\'t define any plan',
             default: null
         },
+        dep: {
```
devinivy, line 255: The nodejs, mocha, and ava CLIs all call this option `require`.  Perhaps for consistency we may call it `require` with a short alias of `req` (since `r` and `R` are taken)?

lib/cli.js
```diff
@@ -557,6 +565,14 @@ internals.options = function () {
         options.sourcemaps = true;
     }
 
+    if (options.deps) {
+
+        for (let i = 0; i < options.deps.length; ++i) {
+
+            require(options.deps[i]);
```
devinivy, line 572: I think for completeness, as well as consistency with the `--transform` option and with CLIs with comparable functionality such as the nodejs CLI itself, we'll also want to support requiring files by path.  Perhaps something like this:
```suggestion
            const resolved = require.resolve(options.deps[i], { paths: [process.cwd()] });
            if (!resolved) {
                console.error(`Could not resolve module used with --require option: ${options.deps[i]}`);
                process.exit(1);
            }

            require(resolved);
```

devinivy (commented):
I support adding the functionality, and I like the direction you took this to address #1033 👍 Just left a couple notes that I think will be useful to address to make this more in-line with other functionality found in lab and throughout the nodejs ecosystem.

damusix:
@devinivy Noted and agreed. I made the changes requested. Please let me know if all is OK so I can add to `API.md`!

devinivy:
Awesome, yeah looks good to me!

damusix:
@devinivy ready

devinivy (approved):
This looks great, thanks!

PR #1042, opened 2022-04-15
Remove node 8 compat fix

kanongil:
Introduced in https://github.com/hapijs/lab/commit/d39ee5584ea913407abbea140ed141bc3d4cf682.

Seems about time… Especially since it will inadvertently be enabled for node 20+.


devinivy (approved):
Thanks!

Nargonath (approved):

PR #1043, opened 2022-04-16
New --coverage-predicates option

kanongil:
This new option is designed to help improve code coverage for platform specific features, by providing a way to optionally signal when coverage has to be disabled. It was inspired by the unconditional coverage exclusion in https://github.com/hapijs/lab/pull/1036/commits/cfbf0601749f963245f6cc00e61a89eb8e0d06a0 from #1036, added only because it is not covered when running on node v12.

I designed it as a powerful primitive, with a simple implementation footprint of about 10 lines of code. For #1036, it can enable coverage on node 14+, by adding a new `has-nullish` predicate in `.labrc.js` when running on node 14+, and revising the "off" line to: `/*$lab:coverage:off$ $not:has-nullish$*/`.

I considered allowing multiple extra predicates with an implicit `AND` condition, but it would add extra complexity to the implementation. Also, it is still possible for a user to combine multiple predicates, by creating new composite predicates, eg. `node14+win32`.

FYI, I have previously discarded a hapi feature I worked on, only because it could not work on Windows, and adding coverage bypasses would mean that I could not reliably reach 100% coverage on the other platforms. With this feature, it could have been completed.

Note, I haven't yet documented how this is used inside coverage bypass comments since the syntax might still change depending on feedback.

### DIFF
--- a/.labrc.js
+++ b/.labrc.js
@@ -1,0 +1,10 @@
+'use strict';
+
+const Somever = require('@hapi/somever');
+
+module.exports = {
+    'coverage-predicate': [
+        process.platform,
+        Somever.match(process.version, '>=14') && 'has-nullish'
+    ]
+};
--- a/.labrc.js
+++ b/.labrc.js
@@ -3,8 +3,8 @@
 const Somever = require('@hapi/somever');
 
 module.exports = {
-    'coverage-predicate': [
-        process.platform,
-        Somever.match(process.version, '>=14') && 'has-nullish'
-    ]
+    'coverage-predicates': {
+        win32: process.platform === 'win32',
+        'has-nullish': Somever.match(process.version, '>=14')
+    }
 };
--- a/API.md
+++ b/API.md
@@ -267,7 +267,7 @@ $ lab --typescript
 - `--coverage-all` - report coverage for all matched files, not just those tested.
 - `--coverage-flat` - do not perform a recursive find of files for coverage report. Requires `--coverage-all`
 - `--coverage-pattern` - only report coverage for files with the given pattern in the name. Defaults to `pattern`. Requires `--coverage-all`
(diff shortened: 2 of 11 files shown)


kanongil:
Hmm, one thing that could make the feature safer, would be if the predicates are specified as an object with `true / false`. This could would allow lab to verify that all predicates are actually used, and that all embedded predicate values exist. However, this is less compatible with the CLI options syntax.

I guess it could work as type `json`, and it could be specified as `--coverage-predicate.nullish true`.

kanongil:
Rebased, and changes CLI interface to match my previous comment. Note, that I did not add any extra verification at this time. However, it is now possible to add the functionality in a later patch.

devinivy (approved):
This is going to be extremely useful, well-done!

kanongil:
Thanks for the support @devinivy. I assume you approve of the new syntax, as an extra embed inside $$?

I considered different approaches, but ended up with this. I think it is pretty clear, while signalling to a reader that it is a special syntax that is not just a comment.
I briefly considered doing it plain, without the $$ (eg. `/* $lab:coverage:off$ not:has-nullish */`), but it made it feel more like a regular comment. I also tried embedding it inside the existing $$, but that felt cluttered and hard to read.

kanongil:
We might also consider exposing the predicates to the `Lab.script()` in test codes, so they can easily lookup these and apply them to eg. a test `{ skip }`option.

devinivy:
Yeah, I think the syntax is appropriate.  I agree it would be nice to be able to resolve those as part of skips, but I also think we can handle that as an iterative improvement.  One concept that comes to mind is to have a special value that can be resolved when running the tests, e.g.
```js
{ skip: Lab.predicate('not:has-nullish') }
```

Nargonath:
Thanks @kanongil, it definitely looks like a powerful feature. 💪 Well done.

PR #1044, opened 2022-04-23
Revise leak detection to use fresh node globals

### LINKED ISSUE #1022: Revise leak detection
#### Support plan

* *is this issue currently blocking your project?* (yes/no): yes
* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: any
* *module version*: 24.3.2
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): standalone
* *any other relevant information*:

#### What problem are you trying to solve?

The leak detection is a nice feature, but the hardcoded whitelist is a maintenance burden for lab. Additionally it report "leaks" that are introduced in the test tooling. Ie. running a test with coverage check enabled can suddenly report leaks, when a regular run has none.

#### Do you have a new or modified API suggestion to solve the problem?

Instead of the built-in whitelist, we can generate it dynamically from the globals right before the first `require()` of test code.

This should provide stability against newly introduced globals in the node runtime, and at least some recognition of tooling introduced globals.

devinivy:
By using a worker, we are able to pretty efficiently get a fresh list of property names and symbols from node's global context.  I also updated the CI config so that it only tests node v12, v14, and v16.  The next major of lab will support v14, v16, and v18.

Resolves #1022

### DIFF
--- a/.github/workflows/ci-module.yml
+++ b/.github/workflows/ci-module.yml
@@ -8,21 +8,4 @@ on:
 
 jobs:
   test:
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu, windows, macos]
-        node: ['*', '14', '12']
-
-    runs-on: ${{ matrix.os }}-latest
-    name: ${{ matrix.os }} node@${{ matrix.node }}
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
-        with:
-          node-version: ${{ matrix.node }}
-          check-latest: ${{ matrix.node == '*' }}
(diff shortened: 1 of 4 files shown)


cjihrig (approved):
Nice. It would be great to have this type of leak detection in Node's new test runner as well if you're up for it @devinivy 😄

devinivy:
@cjihrig Thanks for the review.  Maybe once all this recent work for hapi lands I'll dive into `internals/test_runner`— In the interim please don't hesitate to steal this approach! 😉 🙏

PR #1047, opened 2022-05-02
Fix coverage ignore flag when ts comments are removed

hueniverse:
When the tsconfig is set to remove comments, it also removes coverage skipping comments. This overrides it for testing only.


Nargonath (approved):

Nargonath:
LGTM, thanks for the PR @hueniverse.

kanongil (approved):

PR #1057, opened 2022-10-13
Enable linting of *.cjs and *.mjs files

### LINKED ISSUE #1053: Lint .mjs (et al)
#### Support plan

* *is this issue currently blocking your project?* (yes/no): yeah, kind of
* *is this issue affecting a production system?* (yes/no): nah

#### Context

* *node version*: v16.x
* *module version*: v25.x
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): hapi app, kind of
* *any other relevant information*:

#### What problem are you trying to solve?

`.mjs` files should be linted, as well as `.js` (and `.cjs`, I guess)

#### Do you have a new or modified API suggestion to solve the problem?

Pass `extensions` into eslint options - happy to PR...

dominykas:
Closes #1053 - I had the code lying around, so I may as well make it public... Happy to close this if this is not useful :)

### DIFF
--- a/test/lint/eslint/esm/.eslintrc.cjs
+++ b/test/lint/eslint/esm/.eslintrc.cjs
@@ -3,5 +3,11 @@
 module.exports = {
     parserOptions: {
         sourceType: 'module'
-    }
+    },
+    overrides: [
+        {
+            files: ['*.cjs'],
+            parserOptions: { sourceType: 'script' }
+        }
+    ]
 };
--- a/lib/linter/index.js
+++ b/lib/linter/index.js
@@ -31,6 +31,10 @@ exports.lint = async function () {
         Hoek.merge(configuration, options, true, false);
     }
 
+    if (!configuration.extensions) {
+        configuration.extensions = ['.js', '.cjs', '.mjs'];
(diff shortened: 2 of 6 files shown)


test/lint/eslint/esm/.eslintrc.cjs
```diff
@@ -3,5 +3,11 @@
 module.exports = {
     parserOptions: {
         sourceType: 'module'
-    }
+    },
+    overrides: [
+        {
+            files: ['*.cjs'],
+            parserOptions: { sourceType: 'script' }
```
dominykas, line 10: This was necessary for the `should lint ESM files` test - now that `cjs` is also being scanned, the test needed an update, but it was also complaining that `use strict` is not necessary, because it was getting parsed as a module. TBH, I didn't have time to properly figure out why - eslint should be doing this by default for `*.cjs` files...

But this does unblock further improvements, I'd hope.

devinivy (approved):

devinivy:
Looks great, thanks again @dominykas!

PR #1059, opened 2022-12-19
Bump @types/node

dominykas:
This addresses the `Subsequent variable declarations must have the same type.` regarding `AbortSignal` failure on Node.js 18.


devinivy (approved):
Thank you!

PR #1060, opened 2023-01-21
chore: bump hoek

Marsup:
Needs https://github.com/hapijs/bossy/pull/83 before merging/releasing.

### DIFF
--- a/bin/_lab
+++ b/bin/_lab
@@ -46,5 +46,5 @@ main();
 
 process.on('exit', async () => {
 
-    await Fs.rmdir('./test_runner', { recursive: true });
+    await Fs.rm('./test_runner', { recursive: true, force: true });
 });
--- a/bin/_lab
+++ b/bin/_lab
@@ -2,9 +2,9 @@
 
 'use strict';
 
+const Fs = require('fs/promises');
 const Util = require('util');
 const Cpr = require('cpr');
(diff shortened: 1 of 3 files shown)


bin/_lab
```diff
@@ -46,6 +46,5 @@ main();
 
 process.on('exit', async () => {
 
-    const rimraf = Util.promisify(Rimraf);
-    await rimraf('./test_runner');
+    await Fs.rmdir('./test_runner', { recursive: true });
```
Nargonath, line 49: Considering `rimraf` is the equivalent of `rm -rf`, shouldn't we use `Fs.rm('./test_runner', { recursive: true, force: true });`?

Per Node docs: https://nodejs.org/docs/latest-v16.x/api/fs.html#fspromisesrmdirpath-options
> To get a behavior similar to the rm -rf Unix command, use [fsPromises.rm()](https://nodejs.org/docs/latest-v16.x/api/fs.html#fspromisesrmpath-options) with options { recursive: true, force: true }.

Marsup: Sure.

Marsup, line 49:
```suggestion
    await Fs.rm('./test_runner', { recursive: true, force: true });
```

Nargonath (approved):

Issue #1065, opened 2023-07-24
Add `.ts` to the default `extensions` for `--lint`


dominykas:
#### Support plan

* *is this issue currently blocking your project?* (yes/no): no
* *is this issue affecting a production system?* (yes/no): yes

#### Context

* *node version*: ~18
* *module version*: latest
* *environment* (e.g. node, browser, native): node
* *used with* (e.g. hapi application, another framework, standalone, ...): all over the place
* *any other relevant information*:

#### What problem are you trying to solve?

Running `lab` with `--lint` does not lint `*.ts` files.

#### Do you have a new or modified API suggestion to solve the problem?

When running `lab` with `--typescript`, `.ts` should be added to the default `extensions`: https://github.com/hapijs/lab/blob/master/lib/linter/index.js#L35

Alt: always add `.ts` - same as `.cjs` / `.mjs`.

Happy to PR!

PR #1069, opened 2023-11-14
Always lint .ts

dominykas:
Closes https://github.com/hapijs/lab/issues/1065

~~I took the simpler approach here - an alternative would be to only add `.ts` when `--typescript` is set, but I figure there's probably no harm in always doing this?~~

Had to hide this behind the `--typescript` flag, because otherwise it reports issues with `*.ts` files available in lab itself - and resolving those needs a new eslint config with typescript parser.


Nargonath (approved):
LGTM, thanks for the PR

PR #1071, opened 2023-12-28
Tighten type checking config

kanongil:
This adds [`noUncheckedIndexedAccess: true`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) and [`exactOptionalPropertyTypes: true`](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) to the type checking config.

This helps ensure that the defined typings can work for projects like [reported here](https://github.com/hapijs/joi/issues/2943), which use these stricter than `strict: true` options.

I also updated the peer dependency requirement to v4.4.0+, which is the oldest version that supports these flags.


kanongil:
The [316d8c1](https://github.com/hapijs/lab/pull/1071/commits/316d8c1b259ab323bafe696317ed05a353743e62) patch can be omitted, if #1070 is merged. Otherwise, two additional codes need to be added to the list to handle the expect.error() lines in https://github.com/hapijs/lab/pull/1071/commits/b801faee84cfb9b919d8f9d22b655fbf901c85eb.

devinivy (approved):
Looks great, thanks for getting this going 👍

Nargonath (approved):

PR #1072, opened 2024-06-23
fix: 🐛 place nice with commonjs

damusix:
Current TS config doesn't allow one to use normal ESM imports because they don't have a default export. This config setup should allow it to play nice with modules that setup DTS files without a default exports (like a number of Hapi modules do)


damusix:
Fixed issue upstream. Unnecessary PR, closing.

PR #1073, opened 2024-07-25
feat: add explicit extension support when using Typescript and ESM

Marsup:
The convention on Typescript when using ESM is to require .js files while actually expecting .ts files to be imported, this fails with current lab as obviously the .js file doesn't exist and the `require.extentions` hook doesn't even get a chance to run.

I inspired myself from ts-node for this (especially https://github.com/TypeStrong/ts-node/blob/main/src/cjs-resolve-hooks.ts), but I don't think I need everything as this is just enough to make it work. This is very specifically tailored for the scenario I'm seeing, I'm not sure that it covers everything out there.


damusix (approved):
noice

devinivy (approved):

kanongil (commented):
This needs to guard against adding the hook multiple times (or for sure only call it once).

Also not a fan of the embedded `require()`, which goes against the hapi style guide.

Marsup:
> This needs to guard against adding the hook multiple times (or for sure only call it once).

This is the CLI execution, it's only ever called once AFAIK, at least in normal conditions.

> Also not a fan of the embedded `require()`, which goes against the hapi style guide.

This file is plastered with requires that are not top-level, I don't mind making it global but I also don't see the point if it's never used. Is it important for you?

PR #1078, opened 2024-10-16
Ignore new TypeScript global leak

dominykas:
Ref:
- https://github.com/microsoft/tslib/pull/270
- https://github.com/microsoft/TypeScript/pull/59767


dominykas:
Thanks!

PR #1087, opened 2026-02-20
fix: guard parent access in `addTsEsmHook` for ESM-to-CJS sub-dependency resolution

### LINKED ISSUE #1086: `@hapi/lab` crashes when TypeScript tests depend on ESM packages
### Runtime

node.js

### Runtime version

22.13.1 (any version >= 22.12.0 with unflagged `require(esm)`)

### Module version

26.0.0

### Last module version without issue

_N/A — the bug has existed since `addTsEsmHook` was introduced, but only manifests on Node 22.12+ where `require(esm)` is unflagged._

### Used with

standalone (reproducible with minimal setup)

### Any other relevant information

Affects **every** project using `lab --typescript` that has a dependency with `"type": "module"` in its `package.json`. Node 22.12+ resolves ESM packages via `require(esm)` natively, which triggers a code path in lab where `parent` is `undefined`.

### What are you trying to achieve or the steps to reproduce?

Run TypeScript tests with `lab --typescript` where the test code imports from an ESM package (a package with `"type": "module"`).

### Minimal reproduction

Create a directory with the following structure:

```
lab-esm-repro/
├── package.json
├── tsconfig.json
├── src/
│   └── math.ts
├── test/
│   └── math.ts
└── node_modules/
    └── esm-dep/
        ├── package.json
        └── index.js
```

**`package.json`**

```json
{
  "name": "lab-esm-repro",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "test": "lab test --typescript"
  },
  "devDependencies": {
    "@hapi/lab": "^26.0.0",
    "@hapi/code": "^9.0.3",
    "typescript": "^5.6.0"
  }
}
```

**`tsconfig.json`**

```json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "target": "ES2022",
    "outDir": "dist",
    "rootDir": ".",
    "strict": true
  }
}
```

**`src/math.ts`**

```typescript
import { add } from 'esm-dep';

export function addOne(n: number): number {
    return add(n, 1);
}
```

**`test/math.ts`**

```typescript
import * as Lab from '@hapi/lab';
import { expect } from '@hapi/code';
import { addOne } from '../src/math.js';

export const lab = Lab.script();
const { describe, it } = lab;

describe('addOne', () => {
    it('adds 1 to a number', () => {
        expect(addOne(4)).to.equal(5);
    });
});
```

**`node_modules/esm-dep/package.json`**

```json
{
  "name": "esm-dep",
  "version": "1.0.0",
  "type": "module",
  "exports": {
    ".": "./index.js"
  }
}
```

**`node_modules/esm-dep/index.js`**

```javascript
export function add(a, b) {
    return a + b;
}
```

Then run:

```bash
npm install
npx lab test --typescript
```

### What was the result you got?

```
TypeError: Cannot read properties of undefined (reading 'filename')
    at Module._resolveFilename (/path/to/node_modules/@hapi/lab/lib/cli.js:92:48)
    ...
```

Lab crashes immediately when Node attempts to resolve the `esm-dep` package.

### What result did you expect?

The test should run and pass:

```
  addOne
    ✔ 1) adds 1 to a number (1 ms)

1 tests complete
```

setthase:
**What:** Adds a null guard for `parent` in `internals.addTsEsmHook()` (`lib/cli.js`).

**Why:** When running TypeScript tests (`--typescript`) in an ESM project that imports from an ESM package with CJS sub-dependencies, the crash path is: `require.extensions['.ts']` → `Module._compile` → `loadESMFromCJS` → ESM translators → CJS sub-dependency resolution → `Module._resolveFilename` with `parent === undefined`. The current code accesses `parent.filename` unconditionally, causing `TypeError: Cannot read properties of undefined (reading 'filename')`.

**Fix:** Guard the `parent` access so that when `parent` is `undefined` (ESM-to-CJS transition context), the hook falls through to the original `_resolveFilename`, which handles the resolution natively.

**Test:** Added `test/cli_typescript_esm_dep/` fixture — a TypeScript ESM test that imports from an ESM package which has a CJS sub-dependency (two-layer structure required to trigger the crash). Without the fix, this crashes. With the fix, it passes.

**Ref:** fixes #1086

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,6 @@
 **/.idea
 
 !test/cli_coverage/node_modules
+!test/cli_typescript_esm_dep/node_modules
 test_runner
 test/cli/test/leaks.js
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -89,7 +89,7 @@ internals.addTsEsmHook = function () {
 
     Module._resolveFilename = function (request, parent, ...rest) {
 
-        if (request.endsWith('.js') && parent && parent.filename && (parent.filename.endsWith('.ts') || parent.filename.endsWith('.tsx'))) {
+        if (request.endsWith('.js') && parent && (parent.filename.endsWith('.ts') || parent.filename.endsWith('.tsx'))) {
             const target = Path.join(parent.path, request);
             const tsEquivalent = `${target.slice(0, -3)}.ts`;
 
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -89,7 +89,7 @@ internals.addTsEsmHook = function () {
 
     Module._resolveFilename = function (request, parent, ...rest) {
 
-        if (request.endsWith('.js') && (parent.filename.endsWith('.ts') || parent.filename.endsWith('.tsx'))) {
+        if (request.endsWith('.js') && parent && parent.filename && (parent.filename.endsWith('.ts') || parent.filename.endsWith('.tsx'))) {
             const target = Path.join(parent.path, request);
             const tsEquivalent = `${target.slice(0, -3)}.ts`;
 
--- a/test/cli_typescript_esm_dep/node_modules/cjs-helper/lib/utils.js
+++ b/test/cli_typescript_esm_dep/node_modules/cjs-helper/lib/utils.js
@@ -1,0 +1,6 @@
+'use strict';
+
+exports.double = function (n) {
+
+    return n * 2;
+};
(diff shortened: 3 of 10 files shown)


damusix:
@kanongil this related to ESM coverage by any chance?

.gitignore
```diff
@@ -15,5 +15,6 @@
 **/.idea
 
 !test/cli_coverage/node_modules
+!test/cli_typescript_esm_dep/node_modules
```
kanongil: Please don't add random stuff here.

kanongil (commented):
Overall this looks good.

setthase:
@kanongil - any idea why the CI tests failed? 🤔

Marsup:
Thanks! It's now published.

